PR #17, opened 2021-09-26
fix: typo in exports declaration

MicLeey:
If you copy over the hardhat config in full (including the exports typo), the config won't be picked up and running `compile` will result in the Solidity compiler not being configured properly.

Caused some confusion in the Discord :)


adilanchian (approved):
Dang this is a GOOD ONE. My b! Thanks :)

PR #18, opened 2021-09-26
Fix typo

pujer:
rando -> random


adilanchian:
Haha this is a good catch! Thanks :)

PR #19, opened 2021-09-26
fix: access `abi` key on contract json

MicLeey:
The wave write example is missing an accessor for the `abi` key on the contract json. This was present in the example before, but not in this one.

### DIFF
--- a/Solidity_And_Smart_Contracts/Section_3/Lesson_4_Call_Deployed_Contract.md
+++ b/Solidity_And_Smart_Contracts/Section_3/Lesson_4_Call_Deployed_Contract.md
@@ -122,7 +122,7 @@ Paste the whole JSON file right there!
 
 Now that you have your file with all your ABI content ready to go, it's time to import it into your `App.js` file. Right under where you imported `App.css` go ahead and import your JSON file like so:
 
-`import waveportal from './utils/WavePortal.json';`
+`import contractABI from './utils/WavePortal.json';`
 
 Now that you have this imported and ready to go, you need actually access the ABI content in your code! If you noticed, the contents of that JSON file you imported as a key called `abi`. Thats exactly what you will be accessing in your code here! Let's take a look at where you are actually using this ABI content:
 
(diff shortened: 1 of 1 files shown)


Solidity_And_Smart_Contracts/Section_3/Lesson_4_Call_Deployed_Contract.md
```diff
@@ -138,7 +138,7 @@ const wave = async () => {
         /*
         * You are defining contractABI right here. Let's change this!
         */
-        const waveportalContract = new ethers.Contract(contractAddress, contractABI, signer);
+        const waveportalContract = new ethers.Contract(contractAddress, contractABI.abi, signer);
```
adilanchian, line 141: So if you look down below, you will notice that we actually change this to waveportal.abi which is accessing this as you would expect. 

Did you find this confusing? If so - what would you change it to?

adilanchian:
Please see my review comment! 

Not sure if this is exactly necessary, but would love to know what you think would be the way to go here?

Solidity_And_Smart_Contracts/Section_3/Lesson_4_Call_Deployed_Contract.md
MicLeey: Ah my bad! I meant to change it in the example in the next section, where the accessor is missing but never changed. The GitHub web editor made me cross-eyed :)
https://github.com/zipeducation/buildspace-projects/pull/19/files#diff-5edc7b70d6278a0766d81c066cbcbffca7c242ff5eaf17ac0121107eee7f0758R187

MicLeey:
Also see my comment - sorry for the whoops!

In general though, I don't usually progress until something works. So sometimes I spend half an hour solving something instead of just scrolling down. Perhaps adding a "This will error for now, but we'll fix it later" can help with that.

adilanchian:
Roger that! 

Would love for you to add that in if you are up for it friend :). Lemme know!

MicLeey:
Good to go!

adilanchian:
thanks again for this <3

PR #37, opened 2021-09-27
Update: Removed extra character at line 300

honzcha:
There was an extra " ` " at line 300, causing issues when copy pasting the module.exports code. Removed.


adilanchian:
AWESOME catch. Thanks so much friend <3

adilanchian:
@all-contributors please add @honzcha  for doc

PR #48, opened 2021-09-27
Typo fix - All mention of currency now Ether

offgridauthor:
Went through solidity lessons and changed any mention of 'ethereum' referring to currency; should be 'ether' or ETH.

### DIFF
--- a/Solidity_And_Smart_Contracts/Section_1/Lesson_1_Welcome.md
+++ b/Solidity_And_Smart_Contracts/Section_1/Lesson_1_Welcome.md
@@ -3,7 +3,7 @@
 
 You've made it — hell yeah! Welcome :). My name is Farza and I’ll be your instructor. This project is for devs that want to get more into crypto tech. All you really need to know going into this: some terminal skills, some javascript, some react.js. You'll pick up the rest along the way.
 
-Few action items for you before moving on:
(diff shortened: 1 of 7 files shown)


adilanchian (changes requested):
I think a better change for this would be `ETH` instead of `Ether` whenever we are talking about currency!

Thoughts?

farzaa:
PR creator is right, ether is correct.

adilanchian:
@farzaa I'm a fan of `ETH` instead of `Ether` in this context - thoughts?

@offgridauthor - Looks like we got some conflicts here ! Let me know if you want me to fix or you can :). Just lemme know!

adilanchian:
@offgridauthor - just checking in! Let me know if you are able to get this in, if not I can make the quick change :)

Thanks!

offgridauthor:
@adilanchian - sorry I missed seeing this. Should be resolved now.

adilanchian:
Thanks @offgridauthor !

We actually want to standardize `ETH` here instead of `Ether`. While both are right, we prefer to go with `ETH`. 

Would you mind making this change for me? Sorry about you having to keep going back haha

Lemme know!

offgridauthor:
Submitting new PR  #100

PR #63, opened 2021-09-30
Correct ETH references. Close #43.

### LINKED ISSUE #43: ✨ Lesson Improvement [NFT_Collection]: <SECTION_#2> - <Lesson_4_Create_A_Contract_That_Mints_NFTs>
### What You Are Changing
In the Transactions subsection there is a sentence that says: _For example, sending someone Ethereum is a transaction because we're changing account balances._. 

I think the correct term here is Ether, not Ethereum, cause ETH or ether is the actual currency of the Ethereum blockchain and therefore is the one that can be transferred among wallets. May be I am being a little picky here, but I think is important for beginners to understand the differences between this key terms

toETH
.

matcom-chacha:
Change the way to reference Ethereum currency
in the NFT_Collection Course. Close #43

### DIFF
--- a/NFT_Collection/Section_2/Lesson_4_Create_A_Contract_That_Mints_NFTs.md
+++ b/NFT_Collection/Section_2/Lesson_4_Create_A_Contract_That_Mints_NFTs.md
@@ -183,7 +183,7 @@ The next step is a testnet which you can think of as like a "staging" environmen
  💳 Transactions
 ------------------------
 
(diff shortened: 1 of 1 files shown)


matcom-chacha:
Went through the entire NFT_Collection Course and it seems like these were the only 2 incorrect mentions.

adilanchian:
You're amazing! Thanks so much for doing this :)

adilanchian:
@all-contributors please add @matcom-chacha for doc

PR #65, opened 2021-09-30
Avoid inline component definitions

JoviDeCroock:
Generally the following is considered a bad practice as we are defining a functional component within the render-scope of the rendering component.

```
const App = () => {
  const renderGreeting = () => <p>hi</p>;
  return renderGreeting();
}

vs

const Greeting = () => <p>hi</p>;

const App = () => {
  return <Greeting />
}
```

This is a very subtile weirdness that has the same effect as using a different "key" prop on every render. The child diffing will look at the type of every vnode and see that the functional identity has changed which will mean that we'll tear the subtree down rather than doing an update on it. destroy + create instead of update on every render.

This issue becomes clearer when we consider the following: https://codesandbox.io/s/dawn-leaf-je143?file=/src/App.js notice the CounterInline always being torn down

### DIFF
--- a/NFT_Collection/Section_4/Lesson_2_Connect_Wallet_To_Web_App.md
+++ b/NFT_Collection/Section_4/Lesson_2_Connect_Wallet_To_Web_App.md
@@ -43,9 +43,9 @@ const App = () => {
 
   return (
 		// blah blah all your html here
-    <button className="cta-button connect-wallet-button">
-      Connect to Wallet
-    </button>
+		<button className="cta-button connect-wallet-button">
+			Connect to Wallet
+		</button>
   );
 };
 
@@ -119,9 +119,9 @@ const App = () => {
 
   return (
 		 // blah blah all your html here
-    <button className="cta-button connect-wallet-button">
-      Connect to Wallet
-    </button>
+		<button className="cta-button connect-wallet-button">
+			Connect to Wallet
+		</button>
   );
 };
 
--- a/NFT_Collection/Section_4/Lesson_2_Connect_Wallet_To_Web_App.md
+++ b/NFT_Collection/Section_4/Lesson_2_Connect_Wallet_To_Web_App.md
@@ -18,12 +18,6 @@ const TWITTER_LINK = `https://twitter.com/${TWITTER_HANDLE}`;
 const OPENSEA_LINK = '';
 const TOTAL_MINT_COUNT = 50;
 
-const WalletNotConnected = () => (
-  <button className="cta-button connect-wallet-button">
-    Connect to Wallet
-  </button>
-);
-
 const App = () => {
 
     const checkIfWalletIsConnected = () => {
@@ -49,6 +43,9 @@ const App = () => {
 
   return (
 		// blah blah all your html here
+    <button className="cta-button connect-wallet-button">
+      Connect to Wallet
+    </button>
   );
 };
 
@@ -78,13 +75,6 @@ const TWITTER_HANDLE = '_buildspace';
 const TWITTER_LINK = `https://twitter.com/${TWITTER_HANDLE}`;
 const OPENSEA_LINK = '';
 const TOTAL_MINT_COUNT = 50;
-
-const WalletNotConnected = () => (
-  <button className="cta-button connect-wallet-button">
-    Connect to Wallet
-  </button>
-);
-
 
 const App = () => {
 
@@ -129,6 +119,9 @@ const App = () => {
 
   return (
 		 // blah blah all your html here
+    <button className="cta-button connect-wallet-button">
+      Connect to Wallet
+    </button>
   );
 };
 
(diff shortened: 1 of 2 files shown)


adilanchian:
This is a **REALLY** cool PR. Thanks for submitting it. Some a couple thoughts around here - 

> Generally the following is considered a bad practice as we are defining a functional component within the render-scope of the rendering component.

^ This is definitely valid and something to be aware of, but given the complexity of this UI does this truly matter in this context? 

I think at the very minimum using something like `useCallback` may actually be a better approach! Caching this and only re-rendering on certain prop changes.

Essentially what you are doing is creating another functional component (is what I'm seeing at least!) Is it worth it for something this small?

Thanks again for the CodePen example! Super cool and I'm really excited to learn more from you about this.

JoviDeCroock:
@adilanchian it teaches a practice that will introduce a footgun for a lot of people in the future I feel 😅 I mean, I don't mind removing it but `useCallback` or `useMemo` are still invalid when it comes to a render context and will pose their own issues when it comes to a child-diffing algo

adilanchian:
@JoviDeCroock 10000% agree here. Definitely don't want to teach people the wrong approach here! Curious if you know of any React docs here that we could also point people to! I think it's important to show people the way, but hopefully they feel curious enough to explore it more and it would be cool to provide some resources for them :).

I am definitely fine with this change, just talking about the ways I usually approach this problem and I would love to understand more of when it makes sense to do something like this _OR_ generate a new component?

Thanks again!

JoviDeCroock:
That's a good question, so there are other valid ways of tackling this and I guess this is more a personal preference kind off thing, let me elaborate on my reasoning.

First let's look at an alternative approach that would work as well:

```
const App = () => {
  const renderNotConnectecd = useCallback(() => {
    const NotConnected = () => ...
    // Doing this in two steps so we don't lose our displayName and it actually has a name for react-refresh/react-devtools
    return NotConnected;
  }, [/* so here's the tricky bit, if we add "connectWallet" here then it still recreates the component as in the sandbox*/])
  
  return renderNotConnectecd()
}
```

A similar solution can be done with `useMemo` allthough it runs into a similar issue as elaborate in the deps array of our `useCallback`, do we really want to drop and recreate the component when a prop changes?

Then when there are no props this would be a valid approach but then we get to responsibilities, do we think this `NotConnected` ui-element is a responsibility of `App` if so we can scope it there but generally I'd think the answer (for me atleast) would be no as it's a separate component managing it's own little dom-elements.

I've been looking around but can't immediately find documentation around this but I can refer to a codebase where this becomes clear, in [Preact](https://github.com/preactjs/preact/blob/master/src/diff/children.js#L112) child diffing you can see that we match an oldVNode on _or_ a key _or_ a type, type refers to our function or a string value like `div`. When every render redefines this type these never matches which means we have to call unmount and remount.

adilanchian:
> do we really want to drop and recreate the component when a prop changes?

So for this this would only make sense if our button was changing in some way right? Essentially this button just needs to be shown or not shown so if wallet is not connected don't render this thing.

> Then when there are no props this would be a valid approach but then we get to responsibilities, do we think this NotConnected ui-element is a responsibility of App if so we can scope it there but generally I'd think the answer (for me atleast) would be no as it's a separate component managing it's own little dom-elements.

This is the biggest thing to me and thats why we came to this approach. The original setup of this is actually something I prefer to commonly do when I just want to place components in more "silo-ed" areas, but I don't see them being big enough to separate out into their own components

I guess what I'm trying to say -> **What would be the _best_ approach for a small project like this without over complicating it.** 

Would it make more sense to throw this button directly into the render method of the component and call it a day?

JoviDeCroock:
> Would it make more sense to throw this button directly into the render method of the component and call it a day?

I guess this is also a very valid solution 😅

NFT_Collection/Section_4/Lesson_2_Connect_Wallet_To_Web_App.md
```diff
@@ -41,14 +41,11 @@ const App = () => {
...
-    <button className="cta-button connect-wallet-button">
-      Connect to Wallet
-    </button>
-  );
-
   return (
 		// blah blah all your html here
+		<button className="cta-button connect-wallet-button">
+			Connect to Wallet
+		</button>
```
adilanchian, line 48: Looks like there is a bit of weird spacing at the beginning here if you can fix please :)

NFT_Collection/Section_4/Lesson_2_Connect_Wallet_To_Web_App.md
```diff
@@ -120,14 +117,11 @@ const App = () => {
...
-    <button className="cta-button connect-wallet-button">
-      Connect to Wallet
-    </button>
-  );
-
   return (
 		 // blah blah all your html here
+		<button className="cta-button connect-wallet-button">
+			Connect to Wallet
+		</button>
```
adilanchian, line 124: Spacing here as well!

NFT_Collection/Section_4/Lesson_3_Create_Button_To_Call_Contract.md
```diff
@@ -69,11 +69,13 @@ The rest of the code should already make sense. It looks sorta like the code we
...
-      Mint NFT
-    </button>
-)
+return (
+  ...
+  <button onClick={askContractToMintNft} className="cta-button connect-wallet-button">
+    Mint NFT
+  </button>
+  ...
+);
```
adilanchian, line 78: For this can you do the following - 

```
return (
 {currentAccount === "" ? (
     <button onClick={connectWallet} className="cta-button connect-wallet-button">
         Connect to Wallet
     </button>
   ) : (
      {/** Add askContractToMintNft Action for the onClick event **/}
      <button onClick={askContractToMintNft} className="cta-button connect-wallet-button">
         Mint NFT
      </button>
    )}
);
```

adilanchian (changes requested):
Added some thoughts here :)

adilanchian:
Hey @JoviDeCroock - 

Let me know if you can get the rest of this in! If not I'll get it done :)

Thanks!

JoviDeCroock:
hey @adilanchian sorry about that, this went out of my mind

adilanchian:
No worries! this looks to have done the trick :). Thanks for this woot woot

adilanchian:
@all-contributors please add @JoviDeCroock for doc

PR #79, opened 2021-09-30
Changed --network localhost to --network rinkeby

joshuanwankwo:
Removed the line that talked about changing localhost to rinkeby because in this course so far, we have not deployed to localhost, we only ran run.js script unlike intro to web3 where we ran both run.js and deploy.js locally.
so people who didn't take the intro to web3 course won't get confused.


adilanchian:
Really good catch!! Thanks for pointing this out

adilanchian:
@all-contributors please add @joshuanwankwo for doc

PR #88, opened 2021-10-03
"We have %d total waves!", totalWaves missing

augurnaut:


adilanchian:
It shall be done. Thank you!

PR #100, opened 2021-10-05
Typo fix - use ETH instead of ether/ethereum for currency

offgridauthor:
Changed "ether" and "ethereum" to ETH when referring to currency.


adilanchian:
@all-contributors please add @offgridauthor for doc

PR #108, opened 2021-10-06
[GIT][MV] renamed all folders to match the course

Espequair:
Okay so this one is a pet peeve of mine.
The folder names on github and the course are off kilter and it bugs me immensely.

It doesn't matter if this isn't accepted, I'll have done something about it at least


letsgitcracking:
I feel the same.

adilanchian:
@Espequair change em up! Pretty straight forward I think :)

Thanks!

adilanchian:
@Espequair - so the _MAIN_ change here is the fact that section 1 was actually section 0. The titles of the files were okay?

adilanchian:
Ya this looks to be the case which is good! This means that the map files shouldn't have to change :)

PR #126, opened 2021-10-11
Update Lesson 5, add deployed() method to deploy script

joselvelez:


joselvelez:
> Does this need to be updated on any other lesson?

Good question. I will keep an eye out as I move through the lessons. This is the only place I've noticed so far.

adilanchian:
@joselvelez looks like we have a little merge conflict here !

Let me know if you are able to resolve this  :). 

Thanks!

adilanchian:
Ahh got it! Yes no worries :). 

Since your change is fairly small you can also try branching off of main and adding your changes back! Definitely a cool time to explore this though if you're interested!

PR #135, opened 2021-10-13
Remove unnecessary text from Section 1 Lesson 1

ngnathan:
The line `Call the variable whatever you want.` should be removed since there is no variable in the Solidity contract yet.  If it was referring to the contract name itself `WavePortal`, then changing that name would cause issues with the hre run script in the next lesson.


adilanchian:
@ngnathan thanks for the thoughtful response!

I fully agree! Let's keep it out you have it here and we can go from there 🤘

PR #136, opened 2021-10-14
Add a visible diff to understand what was added

sergey-tyan:
During the course I didn't want to just copy and paste the code you showed, I wanted to understand what exactly was added, and having a `diff` should help with that.

But at the same time, I understand that it will be harder for people to just copy and paste the code because they'll have to remove `+` and `-` at the beginning of the lines.


adilanchian:
Yo this is SUPER cool and didn't know this actually existed haha.

Wonder if there is a way we could do diff && javascript

@farzaa @DavidBarrick thoughts on this?

adilanchian:
Ahh I see! 

Ya this is a much larger change for sure considering we would have to change the way we added things for each code block. I'm going to close out this PR and create an issue and assign it to you so we can have a convo about how we want to go about doing this!

PR #143, opened 2021-10-15
Added "In the next lesson"

AlmostEfficient:
Changed 
> Let's run this and see what we get!
to
> In the next lesson, we'll run this and see what we get!

to clear things up as it seems like you need to figure out how to run it.


adilanchian:
Beautiful. Thanks so much for the change!

PR #197, opened 2021-10-22
Remove `ethers` import at Section_2/Lesson_3_Connect_Wallet_To_Web_App

donni106:
- at this point of the project, `ethers` is not used
  and people are getting: 'ethers' is defined but never used
  when trying to run


donni106:
https://discord.com/channels/869321815833575434/897202307236778004/901126909625245776

Bildschirmfoto 2021-10-23 um 02 20 56

adilanchian (changes requested):
Thanks for the screenshot @donni106 !

I'm cool with this change, but I think we should add an explanation when we do use it. Like  "hey don't forget to import this!"

Could you add this addition for me?

PR #204, opened 2021-10-23
Fix console log string to match comics we love

sayf-t:


adilanchian:
Lmaooo All for this change. Thanks :).

adilanchian:
@all-contributors please add @sayf-ismail for doc

PR #230, opened 2021-10-23
Update Lesson_4_Checking_For_Character_NFT.md

0xQueso:
I think it should be currentAccount as there is no currentAddress variable.


adilanchian:
OOOOF. Ya really good catch here :). Thank you!!

PR #236, opened 2021-10-24
Remove isLoading condition from renderContent

kevinmp:
Removed isLoading feature from renderContent as it's only introduced in Section 4. Having it here will cause an error as the component and related are not yet added to the code.


adilanchian:
You're a legend. Thanks for this catch!

adilanchian:
@all-contributors please add @kevinmp for doc

PR #277, opened 2021-10-28
fix: update Arena/index.js to use showToast instead of boss before sh…

santbob:
…owing toast

In Section 4 of the NFT Game project, `showToast` should be used instead of `boss` in the render function in the code which shows the toast when attacked.

### DIFF
--- a/NFT_Game/en/Section_4/Lesson_1_Finishing_Touches_UI.md
+++ b/NFT_Game/en/Section_4/Lesson_1_Finishing_Touches_UI.md
@@ -441,7 +441,7 @@ Let's then add the HTML to render what our toast will look like! Go ahead an add
 return (
   <div className="arena-container">
     {/* Add your toast HTML right here */}
-    {showToast && (
+    {showToast && boss && characterNFT && (
       <div id="toast" className="show">
         <div id="desc">{`💥 ${boss.name} was hit for ${characterNFT.attackDamage}!`}</div>
       </div>
--- a/NFT_Game/en/Section_4/Lesson_1_Finishing_Touches_UI.md
+++ b/NFT_Game/en/Section_4/Lesson_1_Finishing_Touches_UI.md
@@ -441,8 +441,8 @@ Let's then add the HTML to render what our toast will look like! Go ahead an add
 return (
   <div className="arena-container">
     {/* Add your toast HTML right here */}
-    {showToast && boss && characterNFT && (
-      <div id="toast" className="show">
+    {boss && characterNFT && (
+      <div id="toast" className={showToast ? 'show' : ''}>
         <div id="desc">{`💥 ${boss.name} was hit for ${characterNFT.attackDamage}!`}</div>
       </div>
     )}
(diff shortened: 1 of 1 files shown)


santbob:
@adilanchian something need to be done for this PR to be merged?

adilanchian:
Hey @santbob - 

So I think there is a bit of confusion around how this works and maybe thats the change that needs to be made. Let me explain:

1. We check for `boss` object because we are utilizing properties within the `boss` object in our alert and we only want to render that UI if we have boss data

2. If we only render this toast when `showToast` is true, this may ruin our animations as they are all class based here. The property `attackState` is actually the one that is changing our animation states! So as that property changes the toast will show / hide. The problem here is this may not work if we don't render it right off the bat

So what does this mean then? We still need to keep the `boss` object check, but we may need a little text explaining a bit more why this works so people aren't confused. Let me know if this makes sense and if you need some help with coming up with something here!

santbob:
@adilanchian once again, Thank you very much for taking the time to write a detailed explanation. 🙇

I updated the fix to include both `boss` and `characterNFT` before showing the toast

We would still need the `showToast` check without which the Toast will show up on every render of the component and will give a feeling it is not going away ever.

please check the other users who confirmed the issue https://discord.com/channels/869321815833575434/901160038540902401/903374523355787375 

Thanks

NFT_Game/en/Section_4/Lesson_1_Finishing_Touches_UI.md
```diff
@@ -441,7 +441,7 @@ Let's then add the HTML to render what our toast will look like! Go ahead an add
 return (
   <div className="arena-container">
     {/* Add your toast HTML right here */}
-    {boss && (
+    {showToast && boss && characterNFT && (
       <div id="toast" className="show">
         <div id="desc">{`💥 ${boss.name} was hit for ${characterNFT.attackDamage}!`}</div>
       </div>
```
adilanchian, line 447: Ahhh I see what the confusion is here. Looks like this wasn't added :( SAD. Okay so -

Like I was saying, we want to control the animation with CSS classes not with the `showToast` flag, because this will only control if it's rendered or not.

If you look at the `Arena.css` file, you will notice a class that says show. Show the one that will go through our animations for the toast. We can take this away as well by using conditional className rendering!

**TL;DR - the fix should look like this:**

```
{boss && characterNFT && (
    <div id="toast" className={showToast ? 'show' : ''}>
      <div id="desc">{`💥 ${boss.name} was hit for ${characterNFT.attackDamage}!`}</div>
    </div>
)}
```

So now if we have both the boss data && characterNFT data we should initially render it, BUT we should watch if `showToast` is true because if it is, we will want to render a different className :)

Let me know if this makes more sense now!

adilanchian (changes requested):

adilanchian:
Also - #288 was just opened :). Will give you both credit for this change!

santbob:
@adilanchian updated the PR per the previous comment. Thanks @Gudabu for the #288 it clarifies things

adilanchian:
Sweet!! This will do it :). Thank you for the change!!

adilanchian:
@all-contributors please add @santbob for doc

adilanchian:
@all-contributors please add @Gudabu for doc

PR #280, opened 2021-10-28
Language Update to the lesson 1 in section 1

laksh13:
Just correction of a simple line


adilanchian:
Hey @laksh13 !

Thanks for the PR :). I actually just committed this change in a PR before yours. Thanks for taking the time to make this contribution though :)

PR #284, opened 2021-11-01
Added a missing line

namanmanchanda09:
Farza mentioned "we minted 4 NFTs to the same wallet...", and terminal output is for 4 NFTs as well but the `deploy.js` had 3 NFTs minted. Added the 4th NFT missing code line.


adilanchian:
Ahh yes good catch! I actually think we should change this everywhere to be 3, but it looks like we have screenshots with 4 transactions and such haha. So alas - here we are 😎

Thanks!

PR #294, opened 2021-11-03
Added the missing Boss HP

namanmanchanda09:
The screenshot of the terminal had the upgraded Boss HP as well but the solidity code was missing it.


adilanchian:
Ooo nice catch (like always!)

Thanks!

PR #323, opened 2021-11-06
Update Lesson_1_Get_Local_Solana_Env_Running.md

farzaa:


adilanchian:
I think we should change this wording up a tad to talk about branching off into the different paths here. 

Will do later today!

PR #324, opened 2021-11-07
Change implementation and test to pass in the user's pubkey that call…

psatyajeet:
# Description

In the `add_gif` method, we set `user_address` on the newly created `ItemStruct` (which is meant to store a link to the GIF and the address of the user who submitted the GIF) to the public key of the `baseAddress`. My understanding is that this `baseAddress` is how Solana looks up the Account that stores state for the application. We want to change implementation and test to use the user's pubkey that called `add_gif` to the `AddGif` `Context` so that we can save that to the `ItemStruct.user_address`
## Changes

* Solana_And_Web3/en/Section_2/Lesson_4_Store_Structs_On_Program.md
  *   Update `pub fn add_gif`to reference the user from `ctx: Context<AddGif>` and change the value for `user_address` in `ItemStruct` to `*user.to_account_info().key` instead of `*base_account.to_account_info().key`
  * Update the test call `await program.rpc.addGif(...` by passing in `user: provider.wallet.publicKey,` in addition to `baseAddress`

### DIFF
--- a/Solana_And_Web3/en/Section_2/Lesson_4_Store_Structs_On_Program.md
+++ b/Solana_And_Web3/en/Section_2/Lesson_4_Store_Structs_On_Program.md
@@ -21,15 +21,16 @@ pub mod myepicproject {
     base_account.total_gifs = 0;
     Ok(())
   }
-	 
-  // The fucntion now accepts a gif_link param from the user.
+
+  // The function now accepts a gif_link param from the user. We also reference the user from the Context
   pub fn add_gif(ctx: Context<AddGif>, gif_link: String) -> ProgramResult {
     let base_account = &mut ctx.accounts.base_account;
-		
+    let user = &mut ctx.accounts.user;
+
 	// Build the struct.
     let item = ItemStruct {
       gif_link: gif_link.to_string(),
-      user_address: *base_account.to_account_info().key,
+      user_address: *user.to_account_info().key,
     };
 		
 	// Add it to the gif_list vector.
@@ -48,10 +49,13 @@ pub struct StartStuffOff<'info> {
   pub system_program: Program <'info, System>,
 }
 
+// Add the signer who calls the AddGif method to the struct so that we can save it
 #[derive(Accounts)]
 pub struct AddGif<'info> {
   #[account(mut)]
   pub base_account: Account<'info, BaseAccount>,
+  #[account(mut)]
+  pub user: Signer<'info>,
 }
 
 // Create a custom struct for us to work with.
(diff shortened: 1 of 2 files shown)


adilanchian:
Hey @psatyajeet - 

Thanks so much for the PR here :). So, I'm going to drop some thoughts, but will let @farzaa add onto this as well.

Solana uses accounts to store data per program - you are indeed right. In the normal instance, you are creating a new account to interact with a program. For example, if you hit the GIF Portal multiple times you would use your user account (as you also stated!)

The problem here is your data is siloed in your account which means you would never be able to see any of the other GIFs that others submitted. That is why in this project we have a `baseAccount` that all users interact with. All the data is stored in that account 

TL;DR - your change here is not incorrect, but for the functionality we are going for in this project, this wouldn't be ideal! Let me know if you have any thoughts or concerns here :)

psatyajeet:
Appreciate the comment @adilanchian ! 

Forgive me if I'm missing something, but I believe this change doesn't affect the `baseAccount` that all users interact with. The `gifList` would still be stored in the `baseAccount` so all users would be pulling the `gifList` from a single `baseAccount`. 

The only change being made here is that the `userAddress` in `ItemStruct` (`gifList` is `Vec<ItemStruct>`) should be set to the user that submitted the GIF rather than the `baseAddress`. 

The problem with the existing functionality is that every single GIF is being associated with the`baseAccount`. For example, if we wanted to add a tip button on every GIF to allow tipping the user that submitted the GIF, the `userAddress` on every entry would be the address of the `baseAccount` so we'd be tipping the `baseAccount` every single time.

ItsJamesMurray:
@psatyajeet this was also very confusing to me as well. I am wanting to show the SOL address of the user that submitted a GIF, but  running `item.userAddress.toString()` as suggested in Section 4 Lesson 1 it is showing the baseAccount for every single GIF regardless of whoever submits the GIF. 

Did you make your proposed change on your end successfully?

psatyajeet:
> @psatyajeet this was also very confusing to me as well. I am wanting to show the SOL address of the user that submitted a GIF, but running `item.userAddress.toString()` as suggested in Section 4 Lesson 1 it is showing the baseAccount for every single GIF regardless of whoever submits the GIF.
> 
> Did you make your proposed change on your end successfully?

I did. Yeah, the issue you're pointing out is the one I was having too. my solana program is here: https://github.com/psatyajeet/buildspace-solana-myepicproject and the React project is here: https://github.com/psatyajeet/sol-footy-gifs

adilanchian:
Thanks for the heads up @psatyajeet !

Ya this makes more sense now. I chatted about it with @farzaa and he is going to do some testing on his end before we merge this in!

Thanks again for the awesome PR :)

farzaa:
@psatyajeet @ItsJamesMurray

This looks correct. Running the updated code on my machine now to make sure it doesn't break anything

farzaa:
10/10 find @psatyajeet. Confirmed on my machine. DM me on Discord and I'll send you $100 in ETH for this epic find!

farzaa:
@all-contributors please add @psatyajeet for doc

PR #327, opened 2021-11-07
Solana course: clean up side effects

holic:
It's good hygiene to clean up after side effects when setting up event listeners. It's not critical in this example because `load` only fires ~once per page load and the `[]` effect dependencies also only triggers ~once, but a similar handler on another event without clean up would result in a memory leak.

Happy to add some notes to this to explain _why_ this is being done the way it is (and/or link out to the React docs). Just want to help set up folks for success in future projects by providing best practices ahead of time :) 

Also happy to simplify, e.g.

```jsx
useEffect(() => {
  window.addEventListener('load', checkIfWalletIsConnected);
  return () => window.removeEventListener('load', checkIfWalletIsConnected);
}, []);
```

### DIFF
--- a/Solana_And_Web3/en/Section_1/Lesson_2_Building_Connect_Wallet_Button.md
+++ b/Solana_And_Web3/en/Section_1/Lesson_2_Building_Connect_Wallet_Button.md
@@ -54,9 +54,11 @@ const App = () => {
    * Phantom Wallet
    */
   useEffect(() => {
-    window.addEventListener('load', async (event) => {
+    const onLoad = async () => {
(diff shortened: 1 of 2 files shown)


adilanchian:
Hey @holic -

Ya I love this. Definitely a miss on our end here. I don't think there is a need for explanation on this, but we shall see if people are really confused.

My goal here is that people are intrigued enough to do a little digging on their own and I think this will provide this!

Thanks again - this is epic!!

adilanchian:
@all-contributors please add @holic  for doc

PR #328, opened 2021-11-07
Solana course: form submit instead of button click

holic:
Another good hygiene thing: teach using form submits instead of button clicks. This is for better usability, keyboard handling, etc. Otherwise, if you type into the input and hit enter, the form does nothing. With a wrapping form + `onSubmit`, we get support for both the button click and the type-and-enter in the input.

Happy to move the `onSubmit` handler to its own method rather than inlining it, if you prefer that style-wise.


adilanchian:
Hey @holic !

Ya I can definitely understand this change - do you feel like the `sendGif` method may need to change as well? Instead of using references in react?

holic:
> do you feel like the `sendGif` method may need to change as well? Instead of using references in react?

Do you mean referencing state variables etc.? If `sendGif` is defined in the component, then this should be fine as-is.

adilanchian:
Oops! Sorry I Was thinking of something else :)

This looks good to go to me!!

Thanks so much for this

PR #331, opened 2021-11-07
Update_on_provider_change

NickGonzalez04:
- To change the provider being set with a defined 'provider' variable. Didn't explain why it needed to be a variable after updating our code.

### DIFF
--- a/Solana_And_Web3/en/Section_2/Lesson_3_Store_Basic_Data.md
+++ b/Solana_And_Web3/en/Section_2/Lesson_3_Store_Basic_Data.md
@@ -130,6 +130,7 @@ const { SystemProgram } = anchor.web3;
 const main = async() => {
   console.log("🚀 Starting test...")
 
+	// Create and set the provider. We set it before but we needed to update it, so that it can communicate with our frontend!
   const provider = anchor.Provider.env();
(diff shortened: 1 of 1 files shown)


Solana_And_Web3/en/Section_2/Lesson_3_Store_Basic_Data.md
```diff
@@ -130,6 +130,7 @@ const { SystemProgram } = anchor.web3;
 const main = async() => {
   console.log("🚀 Starting test...")
 
+	// Create and set the provider. We set it before but we needed to update it, so that it can communicate with our frontend!
```
adilanchian, line 133: It looks like the indentation on this is a tad off!

Can you fix for me please :)?

adilanchian (changes requested):
Good catch!

Left one small comment here :). Thanks!

Solana_And_Web3/en/Section_2/Lesson_3_Store_Basic_Data.md
NickGonzalez04: Done! :0

adilanchian:
POGGG thank you :)

PR #335, opened 2021-11-08
Just some grammar fixes

a19i23:
Quick PR to fix the first section description

### DIFF
--- a/Solana_And_Web3/en/Section_0/Lesson_1_What_Are_We_Building.md
+++ b/Solana_And_Web3/en/Section_0/Lesson_1_What_Are_We_Building.md
@@ -9,7 +9,7 @@ All you really need to know going into this is **some terminal skills, some jav
 Few action items for you before moving on:
 
 1. **Connect your Discord in the project dashboard.** All the good stuff is going to be happening in a secret category in the Discord that you'll only have access to once you connect your Discord.
-2. **Connect your Ethereum wallet in the project dashboard.** We'll airdrop you a NFT once you finish this, and if you don't have a wallet connected we won't be able to airdrop it to you!! **We plan to do NFTs on Solana**, but, for now our whole system is built on Ethereum/Polygon so that's where we'll send the NFTs :).
+2. **Connect your Ethereum wallet in the project dashboard.** We'll airdrop you an NFT once you finish this, and if you don't have a wallet connected we won't be able to airdrop it to you!! **We plan to do NFTs on Solana**, but, for now our whole system is built on Ethereum/Polygon so that's where we'll send the NFTs :).
 3. **Be sure to attend the live kick-off**. If you miss it, no worries. It's not required to attend! The recording will be posted in Discord under **#info-board**.
 
 ### **🛠 The project.**
@@ -18,7 +18,7 @@ So what will we be building here :)? Here's a little video:
 
 [Loom](https://www.loom.com/share/545522add8514005a91ececac5abdab2)
 
-### **💎 Earning a NFT.**
+### **💎 Earning an NFT.**
 
 When you reach the end of this project, you'll be able to earn an NFT.
 


deepak-likes-code (commented):
it should remain 'a NFT'

a19i23:
> it should remain 'a NFT'

Hey thanks for the review but this doesn't sound correct to me. Any explanation on why you say it should remain this way?

I googled around a bit to see if there was something maybe I was missing. I found a post where someone else agrees on this

https://ell.stackexchange.com/questions/282631/an-nft-vs-a-nft

In any case if you guys still disagree then this reference below at the end of line 23 should be changed to be consistent, no?
image

adilanchian:
Hey @a19i23 - 

I'm really happy you attached some resources here haha. I have been debating this for quite some time. 

HMMMM Let me get some other input from the team here as well. If this is the case we would probably want to change every lesson that references this you know?

@emboslicev2 @DavidBarrick @farzaa 

Thoughts?

a19i23:
@adilanchian ok cool sounds good 👍

adilanchian:
@a19i23 - 

I went to Ethereum.org and found that they do use "an NFT" (https://ethereum.org/en/nft/#gatsby-focus-wrapper)

Let's do this. Would you be able to just CTRL + F `a NFT` and change them to `an NFT`? 

Good looking out!

adilanchian (changes requested):
Agreed things should be changed to `an NFT`

Would you be open to changing all references in the app like this?

a19i23:
@adilanchian Ok updated all references of `a NFT` to `an NFT`. Check out the capitalization because that's the only question I thought on a couple of these that were titles.

README.md
```diff
@@ -44,7 +44,7 @@ You have access to all of the content right here in GitHub!
    ### **Co-Working Sessions**
    Get access to multiple sessions of collaborative coding! Here you will be able to ask questions directly to the instructor and others in your cohort! All while hacking along to epic Lo-Fi Tunes.
 
-   ### **Complete A Project. Get A NFT.**
+   ### **Complete A Project. Get An NFT.**
```
a19i23, line 47: Here's one of the places where I'm not sure if it should be `Get An NFT` or `Get an NFT` 🤔

Solana_And_Web3/en/Section_0/Lesson_1_What_Are_We_Building.md
```diff
@@ -18,11 +18,11 @@ So what will we be building here :)? Here's a little video:
 
 [Loom](https://www.loom.com/share/545522add8514005a91ececac5abdab2)
 
-### **💎 Earning a NFT.**
+### **💎 Earning an NFT.**
```
a19i23, line 21: Same comment as above. I just followed what was already there in each case but I think it should be consistent.

a19i23 (commented):
I just wanted to show you the places where I wasn't 100 percent sure if it was correct so wrote a couple comments in those updates

README.md
adilanchian: I think here it should be `Get an NFT`

Solana_And_Web3/en/Section_0/Lesson_1_What_Are_We_Building.md
adilanchian: This is great as is!

adilanchian:
@a19i23 this is epic! Thanks for the changes here 🤘

README.md
a19i23: agreed!

a19i23:
ok all set I think we're good to go!

adilanchian:
Looks amazing :). Thanks again for an amazing contribution!!

adilanchian:
@all-contributors please add @a19i23  for doc

PR #359, opened 2021-11-09
added sudo prefix

phenomenon98:
Added sudo prefix to mocha install and yarn install commands. It throws a bunch of errors without it as we are installing globally

### DIFF
--- a/Solana_And_Web3/en/Section_2/Resources/windows_setup.md
+++ b/Solana_And_Web3/en/Section_2/Resources/windows_setup.md
@@ -152,7 +152,7 @@ Now, go ahead and CONTROL + C to stop the validator. **We're never going to actu
 Mocha is a nice little testing framework to help us test our Solana programs.
 
 ```bash
-npm install -g mocha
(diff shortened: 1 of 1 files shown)


adilanchian:
Hmmm so `sudo` in this context only elevates the shell to admin privileges. What were the errors that you were getting here?

This shouldn't affect anything!

phenomenon98:
image

Throws permission errors

adilanchian:
Roger!

Okay - so I don't think this should be an overall terminal change, but a note that says something along the lines of:

"Note: if you have an issue with permissions try running this command with sudo in front of it!" Or something along those lines

adilanchian (changes requested):
See comment above please :)

adilanchian:
Sup @phenomenon98 !

Just checking in here :)

adilanchian:
Hey @phenomenon98 - 

Going to close out this PR for now as I actually haven't seen any other issues expressed about this! Thanks :)

PR #361, opened 2021-11-09
Closing tag div 

hubdev4:
Closing div tag `<div className={walletAddress...`


adilanchian:
Hi @hubdev4 -

So this is actually incorrect! This is the outermost wrapper for this section, so the very last `</div>` matches up here correctly. 

Thanks for the PR! Going to go ahead and close this for now

PR #369, opened 2021-11-10
clearer explanation of how to add setCharacterNFT to App.js

danicuki:


adilanchian:
Hey @danicuki -

I'm actually not quite sure what you are trying to do with this change 🤔. Can you give me bit more of an explanation here?

danicuki:
I noticed during the project that some people did not realize that `App.js` need to be changed.
The instructions say:
` and make sure to pass it in your Arena component in App.js`, but this phrase may pass unnoticed. And also some people new to React may not know how to pass the property to Arena from App.js, so I thought it would be a good idea to explicit through code example.

adilanchian (changes requested):
Okay got it! Thanks for the clarity here

So I think the way we should go about doing this is pretty much adding a little explanation and a piece of code that shows adding the attribute to the Arena component. Something like this - 

Note - Don't forget to head back to `App.js` and add pass the `setCharacterNFT` property to your `Arena` component like so:

```jsx
<Arena characterNFT={characterNFT} setCharacterNFT={setCharacterNFT} />
```

adilanchian:
^ this would go in place of your change you made. Thoughts?

danicuki:
I will make another PR with this recommendation

PR #372, opened 2021-11-11
clearer explanation of how to add setCharacterNFT to App.js

danicuki:


adilanchian (changes requested):
Hey @danicuki -

Thanks for bringing this into a separate PR!

So I think this should live in the actual content - not as a comment in the code if that makes sense!

adilanchian:
Sup @danicuki !

Just checking in here :)

danicuki:
done

adilanchian (approved):
Poggg thanks for this!

PR #408, opened 2021-11-20
Update Lesson_2_Connect_Wallet_To_Web_App.md

metasucks:
Currently doing the course, and the last code example didn't seem to work. Was it intended for the last example to be slightly broken? If not, here are the two small changes to make it functional.


adilanchian (approved):
Hey @metasucks !

Thanks for the changes here :). Def not intended to be broken haha. MUCH APPRECIATED.

PR #421, opened 2021-11-21
Deploy command updated

nishitchittora:
The command needs to be updated. The file is run.js not deploy.js


adilanchian:
Hey @nishitchittora -

This is actually incorrect! The run script just runs some functions on our contract. The deploy script will actually deploy our contract to our local blockchain :). 

With that - I'm going to close this PR, but thats for looking out!

Issue #439, opened 2021-11-23
Translate into spanish the Section_1-lesson_2 of Solidity_And_Smart_Contracts


jesinnott:
Translate into spanish the Section 1 Lesson 2 of Solidity_And_Smart_Contracts.

I am not able to add the translate label.

adilanchian:
Hey @jesinnott got you !

Thanks for tagging the issue in your PR as well :)

helloitsm3:
Closing this as this is outdated. Feel free to re-open it if the issue persists.

PR #440, opened 2021-11-23
Translate into spanish the section_1/lesson_2 of solidity_smart contracts

jesinnott:
This is the translation into spanish for section 1 / lesson 2 of solidity smart contracts.

Issue #439

### DIFF
--- a/Solidity_And_Smart_Contracts/es/Section_1/Lesson_2_Write_Your_First_Smart_Contract.md
+++ b/Solidity_And_Smart_Contracts/es/Section_1/Lesson_2_Write_Your_First_Smart_Contract.md
@@ -1,0 +1,66 @@
+👩‍💻 Vamos a escribir un contrato
+----------------------------
+
+Increíble, Lo hicimos.
+
+Vamos a entrar directamente a nuestro proyecto.
+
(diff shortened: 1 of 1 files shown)


emboslicev2 (approved):

jesinnott:
> @jesinnott this is so amazing to see! You've done an incredible job - We're really fortunate to have you working on this and excited to share it with the world! One thing I'm wondering about - which techincal language do you think needs to be directly translated versus left in english? For example, blockchain remains blockchain but smart contracts becomes contrato inteligente... also do you think it's necessary to have "Hardhat" with a mayúsculo or should all of the pronouns be changed to minúsculo como en español? Curious to hear what you think! Thanks again :)

@emboslicev2 thanks for your feedback! and sorry for the late response I've been busy with my job these days. That's a great question that actually I had too. I would say I have no preferences but it's probably better to be consistent between keep all the tech words in english or translated those to spanish. Looking at the Lesson 0 and which is already translated and seems like blockchain is `cadena de bloques` and smart contract is `contrato inteligente`. So, we can keep the same or update in all places. About hardhat I think we can write it in lowercase. Sorry for not bringing answers but is up to you. Happy to update update the PRs to include these changes if needed. I love to have these conversations now that we are just starting. Have a nice day!

PR #447, opened 2021-11-24
fix: file extensions to .jsx so replit works

MatthewBenjamin:


adilanchian:
Hey @MatthewBenjamin - 

Thanks for this PR :). So we have this weird state were the initial files for non-replit folks has just `.js` instead of `.jsx`. I'm thinking for this section we should just add a little note saying, don't forget if you are on Replit to us `.jsx` files! I called this out at some point in the project, but never hurts to reiterate if thats the way

adilanchian:
Going to close this PR for now! Feel free to re-open it if you feel strongly about it!

PR #477, opened 2021-11-29
added jsx instead of js

ArpitIngle:
added jsx instead of js because js shows error in replilt


ArpitIngle:
> Hey @ArpitIngle !
> 
> Thanks for the changes here :). So we are in a bit of a weird spot right now because the Replit requires us to use `.jsx` files and the Github cloned version, does not! Do you remember seeing a reference to this earlier in the project if you are using Replit that you will be using `.jsx` files?
> 
> Let me know what you think and let's see if we can make this more clear!

yeah , you are right

adilanchian:
Thanks for the reply friend!

Cool - I will go ahead and close this PR for now, but feel free to open another one if you have a better solution for this!

PR #479, opened 2021-11-30
changed 2 to 3 in the js code

ArpitIngle:


adilanchian:
Hey @ArpitIngle !

Thanks for the PR :). So this actually won't work :o. This project only has three different NFT available to mint (in our example at least)

Going to close this out for now, but thansk!

PR #485, opened 2021-12-02
Add update PATH instructions

TylerWoolcott:
Added update PATH instructions after installing Solana before running ``solana --version`` in terminal to avoid ``zsh: command not found: solana``


adilanchian:
Hey @TylerWoolcott !

Thanks for the change request here :). When you were installing this, did it not instruct you to add this export at some point 🤔

Usually it does!

TylerWoolcott:
Hi @adilanchian, 

Yes there is a subtle mention to add the export in the terminal but I wasn't looking for it as it's not mentioned in the docs even though it won't work otherwise and it's not easy to see even once aware. In addition, as I was closely following the 'Setting up Solana on a M1 macOS Machine' docs and it isn't mentioned I didn't think to look for it / do it. Plus I just got finished troubleshooting the 'Having problems with greadlink' issue on my machine so wasn't expecting to get an error running ``solana --version`` and didn't know if the error was connected to my ``greedlink`` error or if it was a different error. It took me a long time to realise what was going on and want to save others who might encounter this the time and confusion.

adilanchian:
Hey @TylerWoolcott !

Thanks for the info here :). Yup this is a good add for sure based on your troubleshooting. I appreciate the steps you took!

Please see my review comments :)

Solana_And_Web3/en/Section_2/Resources/m1_setup.md
```diff
@@ -106,7 +106,7 @@ rustup toolchain install stable
 
 </details>
 
-This might take some time, so don't be alarmed! Once you're done installing, run this to make sure everything is in working order:
+This might take some time, so don't be alarmed! Once you're done installing, you may need to ensure it is in your path by pasting in your terminal ``export PATH="/Users/<your_username_here>/dev/solana"/bin:"$PATH"`` and replacing ``<your_username_here>``, then run this to make sure everything is in working order:
```
adilanchian, line 109: So in solana cargo repo, their log statement actually looks like this - 

`echo "export PATH=\"$installDir\"/bin:\"\$PATH\""`

I think we should change you `<your_username_here>` with `<install_path>` thoughts?

adilanchian (changes requested):

Solana_And_Web3/en/Section_2/Resources/m1_setup.md
TylerWoolcott: Hi @adilanchian, 

Great and yes agreed ``<install_path>`` is more in keeping with their original log statement.

adilanchian: Thanks for the change here!

Sorry - I should have been a little more clear here, I think it should actually look like this:

`PATH="< install_path >/dev/solana"/bin:"$PATH"`

Let me know your thoughts!! Just in the chance that someone actually installs this somewhere else you know?

adilanchian: Also looks like there are a few conflicts ! If you need some help resolving those, lemme know!

TylerWoolcott:
Thanks @adilanchian, 

Yes good point, I made those changes, resolved the conflict and pushed a new commit (hash # 05896d7)

adilanchian:
Thanks for the changes here - really appreciate it!!

PR #489, opened 2021-12-03
Update m1_setup.md: install anchor/web3.js packages globally

twilkinson573:
Earlier on in this guide we install the mocha npm package globally 
```
$ npm install -g mocha
```

It seems appropriate to me in this context that we would also install the anchor & solana/web3.js npm packages globally

```
$ npm install -g @project-serum/anchor @solana/web3.js
```

---------------
Potentially related - the first time I ran through this guide I ran into some difficult to debug opaque issues running the `anchor test...` step that cleared up when I uninstalled and restarted the process using a global npm install

Not sure if these things were related but in hindsight it could have potentially been because I local-installed the anchor npm package in a different directory to where I was trying to use it (ie. in the `myepicproject` directory)

Hope this helps - love what you guys are building! 😄

### DIFF
--- a/Solana_And_Web3/en/Section_2/Resources/m1_setup.md
+++ b/Solana_And_Web3/en/Section_2/Resources/m1_setup.md
@@ -201,7 +201,7 @@ If you got that working, nice, you have Anchor!!
 We'll also use Anchor's npm module and Solana Web3 JS — these both will help us connect our web app to our Solana program!
 
 ```bash
-npm install @project-serum/anchor @solana/web3.js
+npm install -g @project-serum/anchor @solana/web3.js
 ```
 
 ### 🏃‍♂️ Create a test project and run it.


adilanchian:
Hey @twilkinson573 !

Thanks so much for this!! Ya this is super valid. I think I will need to do a tad bit of testing on this...Ultimately, we just use these modules within our web app and no where else. So installing them globally seems like it wouldn't be the right approach here (unless I am forgetting something from the project)

Either way - will definitely give this a quick once over and get back to you soon! 

CC @farzaa for any input

twilkinson573:
Hey Alec, yeah totally see where you're coming from mate - the global install is indeed a bit of a sledgehammer fix!

In that case potentially adding a little more context to signpost where specifically we want to install these npm modules might be the cleaner move (especially since students are bouncing around directories a bit with the Solana install etc!) 👍

adilanchian:
Hey @twilkinson573 !

Thanks for the explanation here :). You definitely are on the right track here! Will get to some testing on this sometime this week!

If you don't hear anything from me within the next 2 days can you just ping me on Discord or `@` me here? Sometimes time gets ahead of me haha

NickGonzalez04:
Hey @twilkinson573 good point! I am testing some things out for installs on M1 and will give this a test as well! From my understanding we might not need `npm i @project-serum/anchor` since we are build its from source. Will confirm that. We still need the Solana/Web3 though.

@adilanchian Will give this a check too!

adilanchian:
Dope thanks @NickGonzalez04 !

adilanchian:
Hey @NickGonzalez04 -

Was curious if you got a chance to actually look at this :)?

adilanchian:
Closing do to inactivity !

PR #509, opened 2021-12-04
Fix type from "assests" to "assets"

TomLouwagie:


adilanchian:
Hey @TomLouwagie !

Thanks for the catch here!

PR #527, opened 2021-12-04
Add a quick note about the "User Rejected Request" error

marcinsola:
I saw that some folks were confused about the 4001 error that is popping up in the console. I was as well, so I decided to add a quick note just to ensure others that this is expected at this stage.


adilanchian:
Hey @marcinsola !

Thanks for this change, I think it's definitely a good one to have considering it's a scary red error haha. I think we just need a _tad_ more explanation behind this though. I think something that says why it's there in the first place!

Let me know and I can give some thoughts if needed :)

marcinsola:
Hey @adilanchian, thanks for the suggestions ;)

I added a statement that the error is expected at this point and explain why that's the case. Let me know if you want me to add anything else there ;)

adilanchian:
Looks dope!! Thanks so much for that :)

PR #528, opened 2021-12-04
Improve NFT images loading time

zeosamaster:
`fetchHashTable` resorts to `MetadataProgram.getProgramAccounts` to get all accounts and that takes a lot of time to respond.

To improve the loading time a bit, we can request the NFT metadata in parallel (as opposed to sequentially).
Instead of waiting for the first NFT's metadata to be loaded before requesting the second NFT's metadata (this creates a bottleneck, considering we'll have 3 sequential requests), we request all the metadata at once and then wait until we get it all back to store all the images in the state in one go.


Solana_NFTs/en/Section_3/Lesson_2_Create_Mint_NFT_Button.md
```diff
@@ -209,17 +209,21 @@ const data = await fetchHashTable(
...
-  }
+    return parse.image;
+  });
+
+  // Wait for all requests to finish
+  const allMints = await Promise.all(requests);
+
+  // Store all the minted image URIs
+  setMints(allMints);
 }
```
adilanchian, line 227: I think this entire function should now be wrapped in a `try / catch` thoughts?

Solana_NFTs/en/Section_3/Lesson_3_Build_Loading_States.md
```diff
@@ -106,17 +106,16 @@ const data = await fetchHashTable(
...
+  const requests = data.map((mint) => {
+    const response = await fetch(mint.data.uri);
+    const parse = await response.json();
+
+    return parse.image;
+  });
+
+  const allMints = await Promise.all(requests);
+  setMints(allMints);
+}
```
adilanchian, line 118: Add the `try/catch`

adilanchian (changes requested):
I like this change a lot! Makes a lot of sense :)

I left a few comments to see your thoughts!

Solana_NFTs/en/Section_3/Lesson_2_Create_Mint_NFT_Button.md
zeosamaster: Given that there was no error handling previously, I didn't want to include too much of a change in one go (to avoid confusing what the actual introduced change was).
But totally agree with that, gonna cook that up when I get the time, thanks for the suggestion 👍

adilanchian: ya thats all good! Would be amazing :) thank you!

zeosamaster: Done, sorry for taking so long!
If any request fails, it'll simply disregard that mint and proceed to the remaining. Any failed request will be filtered out.
Let me know if it looks good :)

Solana_NFTs/en/Section_3/Lesson_2_Create_Mint_NFT_Button.md
```diff
@@ -209,17 +209,29 @@ const data = await fetchHashTable(
...
+    try {
+      const response = await fetch(mint.data.uri);
+      const parse = await response.json();
+      console.log("Past Minted NFT", mint)
+
+      // Get image URI
+      return parse.image;
+    } catch(e) {
+      // If any request fails, we'll just disregard it and carry on
+      return null;
```
adilanchian, line 223: So here I would say we probably want to log out the error right? So we know what failed and why!

Also - does this map not throw an error right now? In your callback method, you are using the await keyword, but you don't define it as async

Solana_NFTs/en/Section_3/Lesson_2_Create_Mint_NFT_Button.md
```diff
@@ -209,17 +209,29 @@ const data = await fetchHashTable(
...
     }
-  }
+  });
+
+  // Wait for all requests to finish
+  const allMints = await Promise.all(requests);
+
+  // Filter requests that failed
+  const filteredMints = allMints.filter(mint => mint !== null);
```
adilanchian, line 231: I also don't believe this will work based on this:

> It rejects immediately upon any of the input promises rejecting or non-promises throwing an error, and will reject with this first rejection message / error.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all

adilanchian (changes requested):
Left some thoughts here! I think there are a few bugs in this actually

Solana_NFTs/en/Section_3/Lesson_2_Create_Mint_NFT_Button.md
zeosamaster: That actually works. The error is caught on the try/catch, it's not spilling to the Promise.all.
Any failed requests will be caught, handled and resolve successfully.

Solana_NFTs/en/Section_3/Lesson_2_Create_Mint_NFT_Button.md
zeosamaster: Right, well caught 🤦 
Adding the async and logging the error 👌

zeosamaster:
Changes pushed, thanks for catching the async error 🙏

Solana_NFTs/en/Section_3/Lesson_3_Build_Loading_States.md
```diff
@@ -106,17 +106,22 @@ const data = await fetchHashTable(
...
+      console.error("Failed retrieving Minted NFT", mint);
+      return null;
     }
-  }
+  });
+
+  const allMints = await Promise.all(requests);
+  const filteredMints = allMints.filter(mint => mint !== null);
+  setMints(filteredMints);
+}
```
adilanchian, line 124: This needs to be updated properly right?

adilanchian (changes requested):
Sorry for the late reply here!!

Looks good other than that one section I pointed out :)

Solana_NFTs/en/Section_3/Lesson_3_Build_Loading_States.md
zeosamaster: You meant the second occurrence in `Lesson_2_Create_Mint_NFT_Button.md`, right? Forgot there were two occurrences in that file, sorry about that. Updated it now.

Sorry for the mess, I think I got it right now 😅

Solana_NFTs/en/Section_3/Lesson_2_Create_Mint_NFT_Button.md
```diff
@@ -209,17 +209,30 @@ const data = await fetchHashTable(
...
-  for (const mint of data) {
+  const requests = data.map(async (mint) => {
     // Get URI
-    const response = await fetch(mint.data.uri);
-    const parse = await response.json();
-    console.log("Past Minted NFT", mint)
-
-    // Get image URI
-    if (!mints.find((mint) => mint === parse.image)) {
-      setMints((prevState) => [...prevState, parse.image]);
```
adilanchian, line 220: Hey don't we still need this? 

Everytime you reload the page this will actually duplicate all the NFTs that were minted!

adilanchian (changes requested):
Sorry for the late reply here - it's been quite busy!!

So I just found a bug actually - lemme know what you think

Solana_NFTs/en/Section_3/Lesson_2_Create_Mint_NFT_Button.md
zeosamaster: That isn't needed here, since with the new approach the whole state is getting set with `setMints(filteredMints)`. You don't need to look at the previous state to add an image (neither check if it's already there). All the images are unique and you can set the whole collection at once.

To illustrate this, imagine you have NFT1, NFT2 and NFT3.
With the previous approach, you're fetching the hashtable, getting the 3 NFTs and then requesting NFT1's metadata, checking if the state already has that image and storing it along all other currently stored images, then moving on to NFT2 and NFT3.
With the new approach, you're fetching the same hashtable and then fetching the metadata for all 3 NFTs at once. When you get them all back (after `Promise.all`), you have a list of 3 images for the 3 minted NFTs and that list completely replaces your previous state, no matter what images you previously had stored. You'll end up with 3 images on the UI.
The only way you get duplicated images on the UI is if you have multiple NFTs with the same image. If you think that's a problem, I can filter out duplicates before setting the state, no problem. In my opinion that'll be confusing because you'll have less images on the UI than minted NFTs. Let me know what you think.

As for reloading the page, I didn't understand what you meant. Reloading the page triggers the requests again and you'll have a clean state and set the whole image list nonetheless, there's no duplication.

adilanchian:
Hey @zeosamaster -

Thanks so much for working with me here and sorry for the late reply haha. Ya great explanation! I completely missed out you were setting the state :).

I say lets go ahead and give it ago! Thanks again for the time

adilanchian:
ALSO - CONGRATS ON YOUR FIRST CONTRIBUTION

✨✨

adilanchian:
@all-contributors please add @zeosamaster  for doc

PR #535, opened 2021-12-05
Added alternative way to get solana airdrop

laitsky:
Hey, I encountered "balance unchanged" problems while trying to get the airdrop — and after reading the solana docs I present alternative solution to get the sol airdrop on the devnet.

### DIFF
--- a/Solana_NFTs/en/Section_2/Lesson_3_Depoy_NFTs_To_Devnet.md
+++ b/Solana_NFTs/en/Section_2/Lesson_3_Depoy_NFTs_To_Devnet.md
@@ -48,6 +48,10 @@ solana airdrop 5
 
 After that's done, you can run `solana balance` again and bam you'll have some SOL. *Note: if you ever run out of fake SOL, you can just run this command again.*
 
+*Note 2: If you encountered **"balance unchanged"** problems while trying to get the airdrop, you can use this command instead:*
+```plaintext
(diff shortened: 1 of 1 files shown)


adilanchian:
Hey @laitsky !

Thanks for this addition! So, you ran into this problem before? Did you verify that you were on the devnet connection? Curious if that was the issue? This should be a a really straight forward command and don't think we necessarily need this, but want to run it by you first to see what you tried 🤔

laitsky:
Hey @adilanchian ! Thanks for the response.

I did verify that my solana local configuration is in `devnet` connection by running the `solana config get` command. My experience that is after running the airdrop command with the `--url` flag for the first time, I don't need to specify the flag again for the subsequent airdrop command for getting the new airdrop, so I'm pretty sure that my machine was running on devnet connection.

It is my first encounter with this problem tho, did the solana-web3 projects and got the airdrop successfully without the `--url` flag. Have no idea why this is happening

adilanchian:
Hey @laitsky !

Thanks for the thorough thoughts here :). This definitely does seem odd, I don't think it hurts to have this change in there, but I also feel like it's just really not needed. If you can do me a favor and keep an eye out for others having the problem in the community, that would be sick!! I will as well and if there is a lot of people running into this we can reopen this PR

Thanks so much for the time!!

PR #548, opened 2021-12-06
in tokenURI() removed calls to bytes() and string()

jester7:
In the code block sections describing the tokenURI function MyEpicGame.sol, it unnecessarily wraps the call to abi.encodePacked() inside a bytes(string()) call before passing to Base64.encode(). The Base64.encode function takes data type bytes which encodePacked already returns.

That particular section is repeated in two code blocks in this lesson.

### DIFF
--- a/NFT_Game/en/Section_1/Lesson_4_Mint_Your_NFT_Locally.md
+++ b/NFT_Game/en/Section_1/Lesson_4_Mint_Your_NFT_Locally.md
@@ -303,19 +303,15 @@ function tokenURI(uint256 _tokenId) public view override returns (string memory)
   string memory strAttackDamage = Strings.toString(charAttributes.attackDamage);
 
   string memory json = Base64.encode(
-    bytes(
-      string(
-        abi.encodePacked(
-          '{"name": "',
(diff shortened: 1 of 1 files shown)


adilanchian:
Hey @jester7 !

Sorry for the late reply here - have you tested this out? I think I'll need just a bit of time to test this out real quick!

jester7:
Hey @adilanchian , I tested it both ways, using the suggested way in the lessons and then by removing the extra function calls per my suggestion. Here is a link to my project where I removed the wrapped calls: [CryptoFishies.sol](https://github.com/jester7/crypto-fishies/blob/26aabfe5e0edf6472aa04eae349668decb465bcf/contracts/CryptoFishies.sol#L127)

The project compiles and the hosted version on Rinkeby works fine. It's not a huge issue since the lesson code works as is, but I was researching the way encodePacked and Base64.encode worked and it seemed like a good optimization.

adilanchian:
Hey @jester7 !

Sorry for the late reply here - it's been quite busy haha. 

Ya this is sick and thanks for your due diligence here <3. LETS GET IT UP!

PR #565, opened 2021-12-08
Added Windows user path shortcut and renamed file

srps:
Renamed Lesson_3_Depoy_NFTs_To_Devnet.md to Lesson_3_Deploy_NFTs_To_Devnet.md
Added separate Windows Powershell cmdlets


adilanchian:
Hey @srps !

Thanks for the changes here :). So I actually don't think we truly need this! We develop these projects on macOS machines (Unix) and that should be good to go for anyone using WSL. I could understand the Windows commands, but I think it's pretty straight forward for Windows users as well. Trying to cut down on amount of content/words if possible :)

Thanks again for this!

srps:
Hey @adilanchian it makes sense that way. I would still make a note at the beginning for Windows users without WSL to use Powershell and replace the ~ with the $home variable.

PR #584, opened 2021-12-11
Add hint to add LoadingIndicator to Arena/index.js

michaelnetter:
Without adding LoadingIndicator, the user gets a Reference Error and a blank page is shown.


helloitsm3 (commented):
There's already an instruction to ask users to import LoadingIndicator at line 11

adilanchian:
Hey @michaelnetter !

Sorry for the late reply here. Thanks @helloitsm3 for the comment btw!! This is comment is very much correct :). We do actually have a comment on this earlier in the lesson :)

With that said, I'm going to close this out! If you feel strongly about it, feel free to open up another PR with a brief description on why you think the change should be made !

Thanks so much!!

PR #588, opened 2021-12-12
Update Lesson_3_Depoy_NFTs_To_Devnet.md

msaug:
solana airdrop cap is now 2SOL instead of 5


adilanchian:
Hey @msaug !

Sorry for the late reply here. So I just looked this up and I saw `airdrop 1`. Thoughts?

https://docs.solana.com/cli/transfer-tokens

msaug:
Hey @adilanchian,
I remember that i was able to `airdrop 2` just fine last week, but anything above would be rejected

adilanchian:
Roger that! Cool let's give it a try and she what happens!

Thanks for this :)

PR #590, opened 2021-12-13
Improve one time event handling

michaelvogt:
Nothing serious, but I thought it makes sense to use this option to make the code a bit nicer.
Verified with the sample app
More info: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#:~:text=once,removed%20when%20invoked.


michaelvogt:
When this change is approved, I'm happy to make the changes in all other lessons where necessary, too

adilanchian:
Hey @michaelvogt !

Thanks for the change here :). So, I feel like the original code is actually just fine here for this scenario. 

With that said I'm going to close this PR, but please feel free to make PRs for whatever you feel will be an improvement :)

TY

michaelvogt:
Don't think so. It seems our understanding of an improvement is too different.

PR #594, opened 2021-12-15
Add info on manually install ethers package

dnyy:
Ethers is not in the dependencies. Added instructions on how to add it from the packages tab in Replit.


helloitsm3 (approved):
Thank you for adding this, this can be useful for those that encounters this issue in Replit

adilanchian:
hi @dnyy !

Thanks for this :). So I actually added the `ethers` package to the starter project in Replit! I think this is a much easier way to go about doing it instead of writing content for it :). 

Thanks for the PR!

PR #629, opened 2021-12-21
Removed 'can' from the sentence and added a comma to the sentence so that it makes sense 

kuriakinzeng:


nachoiacovino:
Not true, we "can" connect to any users wallet. We connected to ours, but we have the ability to connect to any user wallet.

adilanchian:
Hey @kuriakinzeng !

Thanks for the change here , but I think we will keep it as is for now!!

PR #632, opened 2021-12-21
Update code snippet

dnyy:
Update the code snippet for `4-set-claim-condition.js` so it matches the following screenshot. Saw alot of questions about it in discord but no solution.


nachoiacovino (approved):
great! ty

adilanchian:
Hey @dnyy -

Thanks so much for the change!! Great catch indeed :). 

TY!

PR #647, opened 2021-12-22
Fixes expected output to mirror code output

jeffreyiacono:
Relevant line in `scripts/4-set-claim-conditions.js` is:
```
console.log("✅ Successfully set claim conditions!");
```

Paragraph later "After running node scripts/4-set-claim-condition.js
here's what I get:" shows different output than what you'd get if you
run sample code.


jeffreyiacono:
better fix @ https://github.com/buildspace/buildspace-projects/pull/648

PR #675, opened 2021-12-23
added ethereum instead of solana

ArpitIngle:


nachoiacovino (approved):

adilanchian:
Hey @ArpitIngle -

Thanks for making this :). I actually prefer the `web3` terminology instead of `Ethereum` here :)

TY!

PR #703, opened 2021-12-25
Use async/await syntax for checking if the user has the NFT

kuriakinzeng:


adilanchian:
Hey @kuriakinzeng !

Thanks for creating this PR :). We actually use a lot of non-async syntax in this project so if this change is going to happen we should probably change it in all spots. Thoughts?

kuriakinzeng:
Agreed. I did make the changes for all instances but I submitted them as 3 different PR's. What's the best way for me to combine several PR's?

kuriakinzeng:
This PR is included in PR #718

PR #717, opened 2021-12-27
Use async/await to pull addresses and balances

kuriakinzeng:


adilanchian:
Hey @kuriakinzeng -

Thanks for the PR! We use a lot of promise chaining syntax in this project so I think the only valid reason to keep this around is if we change all the content to be async/await

Someone actually did a similar PR to this! Maybe you both can work together on it 👀

kuriakinzeng:
This PR is included in PR #718

PR #733, opened 2021-12-28
mintCharacterNFTAction(index) was firing on load of the page

hackingthemarkets:
mintCharacterNFTAction() is getting fired on load of the page before it exists. This prevents the character selection screen from appearing. This is fixed later in the lesson, but the error at this point prevents the character screen from loading.


hackingthemarkets:
changed m

PR #734, opened 2021-12-28
Add reminder to import LoadingIndicator in Arena index.jsx

hackingthemarkets:
The LoadingIndicator component was used in the Arena component, but there was no mention of importing it. Added reminder to import it at the top of the file.


adilanchian (changes requested):
Hey @hackingthemarkets -

Sorry for the late reply here :(. 

We will probably want to add this to this section as well right? - `🔁 Adding loading indicators to the Select Character page.`

NickGonzalez04:
> 

Hey @hackingthemarkets any update on your end adding the additional thing Alec suggested ? :)

adilanchian:
Closing do to inactivity !

PR #737, opened 2021-12-30
[#fix] s/js/jsx/

wataruoguchi:
We use `index.jsx` for the ReactJS project. However, it is called as `index.js` in the descriptions. It would cause confusion (possibly). So replacing all `index.js` of the ReactJS project with `index.jsx`.


adilanchian:
Hey @wataruoguchi -

Thanks for the change and sorry for the late reply! So this is actually correct :). If you remember at the beginning of the project we talk about Replit / Local dev here! 

Thanks for the thought here though!!

wataruoguchi:
Thanks for taking a look. Wouldn't it be better though using jsx rather than js for consistency, as all other courses are using jsx?

PR #805, opened 2022-01-13
Nitpicking

developing-coder:


adilanchian:
Haha I think we shall keep this one as is :). TY though!

developing-coder:
`Good ol'` is correct though. Check out the [Wiktionary article](https://en.wiktionary.org/wiki/good_ol%27) or just Google it.

Up to you obviously :)

PR #811, opened 2022-01-15
Changed who stores the data on the blockchain!

hritikb27:
Miners only produce blocks of transactions by solving the algorithms defined.
Here is the explained version of what exactly miners do on "ethereum.org" website: "In decentralized systems like Ethereum, we need to ensure that everyone agrees on the order of transactions. Miners help this happen by solving computationally difficult puzzles to produce blocks, securing the network from attacks." [https://ethereum.org/en/developers/docs/consensus-mechanisms/pow/mining]

Fundamentally, EVM stores the whole history of the blockchain and then Nodes are the ones who keeps the copy of the state of the EVM and keeps the network running.


helloitsm3 (approved):
LGTM

AlmostEfficient (approved):

PR #815, opened 2022-01-17
correcting the statement since %100 would produce 0 to 99 and not 0-100

kayvannj:


AlmostEfficient (approved):
Great catch! Thanks for this :)

AlmostEfficient:
@all-contributors please add @kayvannj for documentation

PR #819, opened 2022-01-19
Minor grammatical improvement

developing-coder:


AlmostEfficient:
I think this is meant to be 

```
In later sections, we will go over how to build out your own RNG into your attacks!
```

developing-coder:
> I think this is meant to be
> 
> ```
> In later sections, we will go over how to build out your own RNG into your attacks!
> ```

Seems to me that my version is also correct, both in terms of grammar and meaning (not a native English speaker so take that with a grain of salt). Will leave it as is for the moment. Thank you for your time though!

PR #823, opened 2022-01-20
Minor readability improvement

developing-coder:


cornelg7 (approved):

AlmostEfficient:
Changes 
```
We should be good to go to try and land an attack on Elon
```
to 
```
We should be **ready** to try and land an attack on Elon
```

Up to you Alec!

AlmostEfficient (approved):

PR #845, opened 2022-01-24
Improve explanation of pragma condition (^).

sicongzhao:
Modify the inaccurate explanation of pragma condition syntax (^).

### DIFF
--- a/NFT_Collection/en/Section_1/Lesson_3_Mint_A_NFT_Locally.md
+++ b/NFT_Collection/en/Section_1/Lesson_3_Mint_A_NFT_Locally.md
@@ -44,7 +44,7 @@ Just a fancy comment.  It's called an "SPDX license identifier", You can read m
 pragma solidity ^0.8.0;
 ```
 
-This is the version of the Solidity compiler we want our contract to use. It basically says "when running this, I only want to use a version 0.8.0 or higher, but not higher than 0.9.0. Note, be sure your compiler is set to 0.8.0 in `hardhat.config.js`.
+This is the version of the Solidity compiler we want our contract to use. It basically says "when running this, I only want to use a Solidity compiler with its version 0.8.0 or higher, but not higher than 0.9.0. Note, be sure your compiler is set accordingly (eg. 0.8.0) in `hardhat.config.js`.
 
 ```solidity
 import "hardhat/console.sol";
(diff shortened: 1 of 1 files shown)


AlmostEfficient (approved):

NFT_Collection/en/Section_1/Lesson_3_Mint_A_NFT_Locally.md
```diff
@@ -44,7 +44,7 @@ Just a fancy comment.  It's called an "SPDX license identifier", You can read m
 pragma solidity ^0.8.0;
 ```
 
-This is the version of the Solidity compiler we want our contract to use. It basically says "when running this, I only want to use version 0.8.0 of the Solidity compiler, nothing lower. Note, be sure your compiler is set to 0.8.0 in `hardhat.config.js`.
+This is the version of the Solidity compiler we want our contract to use. It basically says "when running this, I only want to use version no earlier than 0.8.0 but earlier than 0.9.0 of the Solidity compiler. Note, be sure your compiler is set to 0.8.0 in `hardhat.config.js`.
```
AlmostEfficient, line 47: This could perhaps be changed to 

```
"when running this, I only want to use a version 0.8.0 or higher, but not higher than 0.9.0."
 ```

AlmostEfficient:
Would also help to link to the [relevant part in the docs](https://github.com/buildspace/buildspace-projects/pull/845/files#diff-89f165434c169942deca73439b3c6a3e4037fa7d89fb4b77da079cdad8999d60R47). What do you think @adilanchian?

AlmostEfficient (changes requested):

NFT_Collection/en/Section_1/Lesson_3_Mint_A_NFT_Locally.md
AlmostEfficient: Easier to understand for non-native speakers and closes the quote :)

Great catch btw!

sicongzhao: > This could perhaps be changed to
> 
> ```
> "when running this, I only want to use a version 0.8.0 or higher, but not higher than 0.9.0."
> ```

Yes, I agree. This is a better way to phrase it.

AlmostEfficient (approved):

AlmostEfficient:
@all-contributors please add @sicongzhao for documentation

PR #857, opened 2022-01-29
Added a note

Tarun-k-Gupta:


NFT_Collection/en/Section_1/Lesson_2_Get_Local_Env_Running.md
```diff
@@ -29,6 +29,8 @@ cd epic-nfts
 npm init -y
 npm install --save-dev hardhat
 ```
+Note: If you have hardhat already installed from previous projects, you can omit the last command (npm install --save-dev hardhat).
```
AlmostEfficient, line 32: Hardhat is installed locally, per project. This would only apply if you were somehow setting up in the same directory? 

IMO this just adds confusion. Please close this PR.

Tarun-k-Gupta: Actually, when I did NFT_Collection after completing wave-portal project, I didn't use this command and it worked fine. Also, both the projects were in different directories. 

But still if it creates confusion, I will close this PR.

AlmostEfficient: You're right! It isn't **required**, but it's recommended to avoid version issues. Worth keeping in, I reckon? https://hardhat.org/getting-started/#installation

Tarun-k-Gupta: Oh right, so there might be some version issues.

Cool, I will close this PR now.

PR #863, opened 2022-02-01
Updating .js to .jsx for Replit users

jfed8:
Honestly not a pro in OS PRs, so this can serve as a solid test-run. When using Repl.it shell, running `node createKeyPair.js` will fail. This may not be obvious to all, but you need to run `node createKeyPair.jsx` (because Repl.it uses .jsx files)


helloitsm3:
Thanks for creating a PR for this but if you create a .js file on Replit, you will still be able to run `node createKeyPair.js` without any issues. Just tested it and I can confirm it works. You can close this PR now @adilanchian. 

image

PR #871, opened 2022-02-03
Update Lesson_1_Get_Your_Local_Ethereum_Network_Running.md

constanteen:
Tried running on my system with Node 15 and I encountered an error: "You are using a version of Node.js that is not supported by Hardhat, and it may work incorrectly, or not work at all."  Did some research and found this useful https://hardhat.org/reference/stability-guarantees.html


AlmostEfficient:
This is great! Can you add a link to the NodeJS releases page (https://nodejs.org/en/about/releases/) so people can find out which version they should be running?

AlmostEfficient (changes requested):
This is great! Can you add a link to the NodeJS releases page (https://nodejs.org/en/about/releases/) so people can find out which version they should be running? Something like

> We recommend running Hardhat using the current LTS Node.js version or you may run into some issues! You can find the current releases [here](https://nodejs.org/en/about/releases/).

constanteen:
Yeah. This looks better. Thanks. I updated it.

AlmostEfficient (approved):

AlmostEfficient:
@all-contributors please add @constanteen for documentation

PR #876, opened 2022-02-05
Update Lesson_3_Compile_Contract_Locally.md

Hakeemmidan:
Add small comments about `process.exit(num)` that hopefully provide more context in a JavaScript code snippet.

### DIFF
--- a/Solidity_And_Smart_Contracts/en/Section_1/Lesson_3_Compile_Contract_Locally.md
+++ b/Solidity_And_Smart_Contracts/en/Section_1/Lesson_3_Compile_Contract_Locally.md
@@ -38,11 +38,12 @@ const main = async () => {
 const runMain = async () => {
   try {
     await main();
-    process.exit(0);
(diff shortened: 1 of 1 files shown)


AlmostEfficient:
I'm not sure if the value these add for the clutter introduced is worth it 🤔

Hakeemmidan:
> I'm not sure if the value these add for the clutter introduced is worth it 🤔

@AlmostEfficient 

I had to look up what each does when I was going through the course because it's not apparent (in my opinion).

One would think that `process.exit(0)` is the error exit since 0 is a falsey value in JavaScript. (I thought that there was a typo  in the code, and that they should be switched up, which prompted me to look `process.exit(<num>)` up.)

So I believe that having the comments might remove some potential confusion.

But feel free to close it if you don't see it adding value at the moment.

AlmostEfficient (approved):

AlmostEfficient:
@all-contributors please add @Hakeemmidan for documentation

PR #893, opened 2022-02-06
Update Lesson_1_Drop_Timer_And_Sold_Out_State.md

ponyolovesham:


adilanchian:
@ponyolovesham sup!

Can you talk more about your change here?

ponyolovesham:
@adilanchian hi!
running `ts-node ~/metaplex-foundation/metaplex/js/packages/cli/src/candy-machine-cli.ts update_candy_machine -e devnet -k ~/.config/solana/devnet.json -cp config.json` to update the cache throws an error, this script is from v1 (v1 contents were deprecated but this script somehow made it through) so we updated the script to point to the v2 script, which i tested and confirm worked

second changes are just semantics, changes will further clarify their format (directory) visually as well

third change is a continuation of updates where i think calling `goLiveDate` as `goLiveData` is anti-pattern, where every other properties we're indexing out of the `candyMachine` keeps the naming the same

PR #904, opened 2022-02-11
Small nit on Section 2 Lesson 2 (JS DAO)

alexisweill:


adilanchian:
Another good catch - thank you!!

PR #908, opened 2022-02-15
update the update_candy_machine to use v2 command

pjerryhu:
see title


AlmostEfficient:
Nice catch! Looks like I missed that file :)

AlmostEfficient (approved):

PR #912, opened 2022-02-17
front/back end changes to attackComplete event

scott-a-m:
Changes proposed for both solidity contract (section 2.3) and front-end (section 3.6) code for the `attackComplete` event. Added `msg.sender` as a parameter to be emitted in the event so that we can check on the front end whether the attack has come from our own player or not.

### DIFF
--- a/NFT_Game/en/Section_2/Lesson_3_Adding_Utility_Functions.md
+++ b/NFT_Game/en/Section_2/Lesson_3_Adding_Utility_Functions.md
@@ -4,7 +4,7 @@ Below I'm going to be going over a few functions. These won't seem really useful
 
 We need a way to check if a user has a character NFT we've given them, and then retrieve the NFT's attributes if the NFT exists. Why?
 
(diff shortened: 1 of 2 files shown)


AlmostEfficient:
Hey Scott! Thanks a bunch for this. Could you change the wording here to be more consistent with the rest of the project?

> Notice how we also passed `msg.sender` as a parameter so that we can check on the front-end later whether the attack has come from our address or someone else's.

> which we will then use to update the state of both our boss and character NFT (or just the boss if the attacking character isn't ours).

scott-a-m:
No problem. I've made some further changes. Hopefully the style is okay now. If still not, let me know what to type and will finalise it for you :-)

scott-a-m:
Sorry, just checked files and it seems my IDE has reformatted the pages, let me commit the files again....

scott-a-m:
Okay, changes have now been made and previous formatting restored. Please give it a look :-)

PR #916, opened 2022-02-18
Fix metaplex path 

### LINKED ISSUE #915: ✨ [Solana_NFTs]: <SECTION_2> - <LESSON 3> Wrong path for metaplex
### What You Are Changing
Problem: 
- on line `92`, `132`, and `158`. The path for metaplex is `~/metaplex-foundation/metaplex/..`, this caused error because the folder is called `metaplex` and doesn't start with `metaplex-foundation`

Solution:
- Remove `metaplex-foundation` from the path.

jte0711:
Fix #915

### DIFF
--- a/Solana_NFTs/en/Section_2/Lesson_3_Depoy_NFTs_To_Devnet.md
+++ b/Solana_NFTs/en/Section_2/Lesson_3_Depoy_NFTs_To_Devnet.md
@@ -89,7 +89,7 @@ Now we're going to use Metaplex's `upload` command to upload our NFTs that live 
 Notice how we do `./assets` in the command below. That means we need to run this command from just one level outside of the `assets` folder.
 
 ```plaintext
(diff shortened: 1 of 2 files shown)


AlmostEfficient (changes requested):
Hi, there's about 20 other instances of us using `metaplex-foundation` instead of `metaplex`. 

That is the original structure. Since it doesn't make a difference, let's change all relevant instances to `metaplex`. Please update the remaining ones. 

Note - there's a few links, pls don't change those, they will break

jte0711:
Hi @AlmostEfficient , I changed all the local paths (excluding the URLs). Please check it again. Thanks

AlmostEfficient (approved):

AlmostEfficient:
Thank you! 

@all-contributors please add @jte0711 for documentation

PR #935, opened 2022-02-20
Simplify json in Polygon_ENS/en/Section_1/Lesson_4_Sell_Domains_As_NFTs.md

aeither:
abi.encodePacked return in bytes. The code convert to string then to convert back to bytes making the code looks more complicated. 
bytes and string convertions are removed. the json's value remain the same.


AlmostEfficient (approved):

AlmostEfficient:
@all-contributors please add @aeither for documentation

PR #939, opened 2022-02-20
Fix code issue in Section2 Lesson3 mintDomain

koji:
The current code is missing `await`

```js
tx = contract.setRecord(domain, record);
await tx.wait();
```

```js
tx = await contract.setRecord(domain, record);
await tx.wait();
```


AlmostEfficient:
Looks like this has already been fixed upstream by another PR, thanks a bunch!

PR #931, opened 2022-02-19
Polygon_ENS: Section_2 - Lesson_3, Section_3 - Lesson_2, Section_4- Lesson 2

altugbakan:
### What You Are Changing
Hello! I have fixed some spacing errors and missing `await` operators. Also, I think it should be `contractAbi` instead of `contractABI` since it does not compile otherwise.

### Anything else?
Love the courses! Please keep it up.

PR #947, opened 2022-02-20
changed contractABI and added await

Kn0wn-Un:
ABI file is imported as contractABI but contractAbi is used in function
not adding await to tx = contract.setRecord(domain, record); causes error in await tx.wait();


Kn0wn-Un:
Yes, this has been fixed in #931 closing this

PR #954, opened 2022-02-21
Update Lesson_1_Deploy_NFT_Bundle.md

ponyolovesham:


AlmostEfficient:
Could you update this and match all other scripts too?

PR #957, opened 2022-02-21
Fixes issue in mintCharacterNFTAction function

jafman:
Fixes issue stopping mintCharacterNFTAction function from triggering.

### DIFF
--- a/NFT_Game/en/Section_3/Lesson_5_Building_Character_Select_Page.md
+++ b/NFT_Game/en/Section_3/Lesson_5_Building_Character_Select_Page.md
@@ -176,7 +176,7 @@ This is amazing, but we can take it one step further - **a one button click to m
 
 ```javascript
 // Actions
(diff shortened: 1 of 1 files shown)


AlmostEfficient:
Hi! Thanks a bunch for this. Could you please add a bit more detail about what the issue was? I can take a guess, I just want to make sure I'm understanding it.

jafman:
`const mintCharacterNFTAction = (characterId) => async () => {` returns a function that is never called.

so the `onclick` event below does not fire.

`<button
        type="button"
        className="character-mint-button"
        onClick={()=> mintCharacterNFTAction(index)}
      >{Mint ${character.name}}</button>`

AlmostEfficient (approved):

AlmostEfficient:
Thank you!

jafman:
> Thank you!

My pleasure ...

PR #958, opened 2022-02-21
Update Lesson_3_Let_Users_Mint_Your_NFT.md

ponyolovesham:
https://dev.to/danialdezfouli/what-s-wrong-with-the-async-function-in-useeffect-4jne :)


AlmostEfficient:
Hi, thanks for contributing! Can you help me understand why you've added the async call for these useEffects? They are not async functions.

ponyolovesham:
> Hi, thanks for contributing! Can you help me understand why you've added the async call for these useEffects? They're not doing anything asynchronously.

hi! passing in an async callback into useEffect is a famous anti-pattern. so therefore the async functions are now being called inside of useEffect, which is the recommended way.
i’m interested in knowing why you think the aync functions aren’t doing anything. every one of them is being invoked. lmk if this isn’t clear

PR #964, opened 2022-02-22
Update Lesson_2_Show_Minted_Domains.md

johnamcconnell:
add info about updating the useState for setLoading as well otherwise it will throw errors

### DIFF
--- a/Polygon_ENS/en/Section_4/Lesson_2_Show_Minted_Domains.md
+++ b/Polygon_ENS/en/Section_4/Lesson_2_Show_Minted_Domains.md
@@ -31,6 +31,9 @@ const updateDomain = async () => {
 
 Nothing new here, it’s just a smaller version of our `mintDomain` function. You’re a pro at this stuff now so I expect you to understand this hehe.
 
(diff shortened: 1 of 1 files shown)


Polygon_ENS/en/Section_4/Lesson_2_Show_Minted_Domains.md
```diff
@@ -31,6 +31,9 @@ const updateDomain = async () => {
 
 Nothing new here, it’s just a smaller version of our `mintDomain` function. You’re a pro at this stuff now so I expect you to understand this hehe.
 
+also remember to add ```const [loading, setLoading] = useState(false);``` at the top of your App.js so that you are setting the useState and your app knows about calling the setLoading in this function
+
+
```
adilanchian, line 36: So I would actually change this to:

Note: Don't forget to add `const [loading, setLoading] = useState(false);` at the top of  `App.js`!

adilanchian (changes requested):
Thanks for the PR!

Just some small tweaks here!

Polygon_ENS/en/Section_4/Lesson_2_Show_Minted_Domains.md
johnamcconnell: sounds good to me! anything to make life easier is always a good thing.

PR #974, opened 2022-02-24
Fixing Async for mintCharacterNFTAction

holyaustin:
Fixing the Async / Await for mintCharacterNFTAction function that would not allow users to mint NFT Characters. i


AlmostEfficient (approved):

AlmostEfficient:
Much cleaner. Thank you!

PR #988, opened 2022-02-28
contractABI changed

holyaustin:
import contractABI from './utils/contractABI.json' gives error because the actual variable name used is contractAbi


AlmostEfficient:
Hello! I think someone else has already had a PR merged for this one. Could you please fetch upstream and let me know if it's still an issue? Thanks!

PR #1002, opened 2022-03-06
Added npm install --save dotenv

XanderCalvert:
I feel this was too vague of what the solution is to not have users to submit their private information about their wallets online. It stated that it will be covered later but didn't state where. Brought the information forward from that location (still there as well) as it isn't unreasonable for people to be already using version control already on this project.


AlmostEfficient (approved):

AlmostEfficient:
Thank you Xander!

PR #1003, opened 2022-03-06
feat: add switch network button back.

moojing:
I just finished the project and I found that there's a small piece of missing code in Section 4 `Show minted domains and update records` step.

Screen Shot 2022-03-07 at 12 50 03 AM

There should be a switch network button because it's just implement in the previous step named `Show the user their wallet and network` in the same section.  Just like this : 

Screen Shot 2022-03-07 at 12 49 16 AM


AlmostEfficient (approved):

AlmostEfficient:
Great catch! Thanks so much!

AlmostEfficient:
@all-contributors please add @moojing for documentation

PR #1023, opened 2022-04-02
Use 'same' instead of 'some' in NFT_GAME S3L4

Alejandro2180:


AlmostEfficient (approved):

AlmostEfficient:
Another one, thanks a bunch!

PR #1041, opened 2022-04-23
Check for chain id only after we have the wallet address 

karacsonybarni:
On Replit if we don't check for the address, we get 
"Uncaught TypeError: Cannot read properties of undefined (reading 'id')"
because the chain is undefined.
Thanks for mehul shah for sharing this fix on Discord!


AlmostEfficient (approved):

AlmostEfficient:
Thank you!

@all-contributors please add @karacsonybarni for documentation

AlmostEfficient:
@all-contributors please add @karacsonybarni for documentation

PR #1045, opened 2022-04-24
front/back-end fix

scott-a-m:
Added `msg.sender` to `attackComplete` event so that we can check whether the attack was made by our player or by another one and update the front-end respectively.


AlmostEfficient:
Thanks for being so resilient Scott! This is definitely an improvement.

AlmostEfficient (approved):

AlmostEfficient:
@all-contributors please add @scott-a-m for documentation

scott-a-m:
> Thanks for being so resilient Scott! This is definitely an improvement.

Thank you very much!

PR #1053, opened 2022-04-28
fixed text error

lodgendury:
rearranged text because it shows abruptly (shows only "The") on the browser.


AlmostEfficient (approved):

AlmostEfficient:
Woah, that's whack! Thanks so much for the fix!

PR #1078, opened 2022-05-11
fixed text error

lodgendury:
code so far only has a function that "returns (address)" ,  "returns (string)" is nowhere in the code at this point.


AlmostEfficient (approved):

AlmostEfficient:
Nice catch, ty!

PR #1089, opened 2022-05-21
Lesson_1_Randomly_Pick_Winner: correcting randomness error in code

goodpilotfish:
Corrected the randomness to 50%. Note: Lesson does not exist in Spanish yet.


helloitsm3:
<=50 is correct. An alternative to doing this is to compare `seed % 2 == 0` where any given number have a 50-50 chance of it being an even number.

genpunk4i:
I believe the original PR is correct. seed % 2 yields 0 or 1. The expression for 50% in this case needs to be, seed % 2 < 1. For x % 100, we can get a number between 0-99. x <= 50 yields 51% not 50%

AlmostEfficient:
Thank you both for contributing, this is awesome!

PR #1096, opened 2022-05-22
Add return to argument checks to avoid [ERR_HTTP_HEADERS_SENT]

0xcen:
Fixes cannot set headers after they are sent to the client error.


AlmostEfficient (approved):

AlmostEfficient:
Oops, looks like I got too sleepy when writing this. Thanks so much for the fix!

AlmostEfficient:
@all-contributors please add @0xcen for documentation

0xcen:
I had a great time doing the project. Thanks for making it! Cheers!

PR #1097, opened 2022-05-22
fix: adjust lessons 3 & 4 in solana-pay-store

ShaunSHamilton:
Mostly typo fixes, with one fix to use `fs/promises` (otherwise there is no reason to await:
1. Change references to `lib/` directory to all be lowercase
2. Change `await fs.writeFileSync` to `await writeFile`


AlmostEfficient (approved):

AlmostEfficient:
Thanks Shaun! This is great!

PR #1099, opened 2022-05-23
Remind of changing contract related data

d-wasilewski:
There's nothing said about changing data on the frontend part after changing things in the contract and deploying. I know it's simple but some people might forget about it:)


AlmostEfficient (approved):

AlmostEfficient:
Great catch, thanks a bunch!

PR #1103, opened 2022-05-24
Update Lesson_2_Show_Minted_Domains.md

6str:
changed comments
    // This will call the updateDomain function we just made
    // If editing is not true, the mint button will be returned instead
to
    {/* This will call the updateDomain function we just made */}
    {/* If editing is not true, the mint button will be returned instead */}

as the comments are inside render functions and show on the page

n.b. the 2nd and 3rd forward slash + asterisk combos in this comment are showing up weird with the asterisks missing


AlmostEfficient (approved):

AlmostEfficient:
Haha this is a great find! Thanks so much!

PR #1104, opened 2022-05-24
Solana Pay: add note to restart after updating .env

mcintyre94:
- Next needs restarting to pick up changes to `.env`, otherwise the code using the new env var won't work
- Fixed a tiny typo on the same line


AlmostEfficient (approved):

AlmostEfficient:
Wooo! Thanks a bunch Callum! I really appreciate it!

PR #1106, opened 2022-05-24
Fix enum key and reference all occurrences

zeosamaster:


AlmostEfficient:
Hmm, I'm not sure this is right. Could you please provide your reference?

https://github.com/solana-labs/wallet-adapter/blob/master/packages/core/base/src/types.ts

AlmostEfficient (approved):

PR #1118, opened 2022-05-31
updated Solana Pay Store for Web3.Storage

maweiche:
removed Pinata instructions and added Web3.Storage instructions.  refactored addProduct component to use Web3.Storage and removed infura.


AlmostEfficient:
Yo I completely forgot about this -- why the close? I can jump back in now and sort it out :)

PR #1132, opened 2022-06-08
Fix a typo in Section 1 Lesson 4

soranature:


AlmostEfficient:
Looks like an upstream PR fixed this. Thanks so much!

PR #1160, opened 2022-06-23
Replaced Rinkeby with Polygon Mumbai

vijethx:
all the references to Rinkeby testnet is replaced with polygon Mumbai
- https://github.com/buildspace/buildspace-projects/issues/1159


AlmostEfficient:
This has been on my radar for a while, thanks so much for this!

PR #1167, opened 2022-06-25
Fixed a spacing typo and set the amount variable to the correct amount (when minting tokens)

Dmigit:
# These are some simple typos that I found while going through the JavaScript DAO Project

## Typo 1
The "amount" constant that stores the amount of ERC-20 tokens we will mint (in line 98) should be set to 1,000,000 instead of 0 to actually mint the tokens and to print the same result in the terminal as the one that is shown in the guide. 

For example, when I followed the instructions verbatim, this is what I got in the terminal:

Screen Shot 2022-06-25 at 10 51 50 AM

However, this is what you're supposed to get according to the guide:

Screen Shot 2022-06-25 at 10 58 07 AM

## Typo 2

This is just a small one. In line 192, there should be a space between "contracts" and "because"

## Conclusion

Thank you for reading this PR. I'm really enjoying the project so far and I hope that I could help make it just a little better.

### DIFF
--- a/JS_DAO/en/Section_3/Lesson_1_Deploy_ERC20_Contract.md
+++ b/JS_DAO/en/Section_3/Lesson_1_Deploy_ERC20_Contract.md
@@ -95,7 +95,7 @@ const token = sdk.getToken("INSERT_TOKEN_ADDRESS");
 (async () => {
   try {
     // What's the max supply you want to set? 1,000,000 is a nice number!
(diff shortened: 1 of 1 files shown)


AlmostEfficient (approved):

AlmostEfficient:
Great suggestion and catch Dylan! Thanks a bunch!

PR #1204, opened 2022-07-03
Paraphrased work in web3 section

thesoorajsingh:
Since the redesign of buildspace, the Work in Web3 Nav section was moved to the header from the side bar, noticed that it wasn't updated in the Mint NFT Collection.


AlmostEfficient:
Hey Sooraj, thanks for this!

We're just going to remove the work in web3 section from all projects, so this PR won't be necessary!

Issue #1207, opened 2022-07-08
A better way to deploy Ethereum Smart Contracts - thirdweb deploy


vijethx:
Discovered this interesting [video](https://www.youtube.com/watch?v=CaODHRnFgBw) by Dhaiwat yesterday. 

This might solve the problem of accidentally committing private keys to Github. It'd be awesome if the deployers are given an option to choose between the current .env file method or this method

helloitsm3:
Closing this as this is outdated. Feel free to re-open it if the issue persists.

PR #1208, opened 2022-07-08
Added thirdweb deploy

vijethx:
Proposing changes with respect to 

- #1207


vijethx:
If this is ok, will add this to all the supported projects

PR #1211, opened 2022-07-08
Fixed render functions and a typo

dylanintech:
# What I Changed
First of all, I liked this project because of how interesting Solana is but also because it really allowed us to express our creativity and be original with, for example, the types of GIFs we used. I just ran into some minor issues in the **Building a GIF Grid** section. I was following along and writing down the render functions but then I realized that my page would not change at all even if my wallet was connected. I was confused as to why this was happening and looked at the guide again to see if I had missed anything. However, after staring at the code for a little longer I saw that the JSX in these functions needed to be wrapped in a return statement. I did this but still, nothing changed, so I tried changing the wrapper of the function from parentheses to brackets, and the page finally changed. Also, I fixed a small typo. I hope this was useful and thanks again BuildSpace!

### DIFF
--- a/Solana_And_Web3/en/Section_1/Lesson_3_Building_GIF_Grid.md
+++ b/Solana_And_Web3/en/Section_1/Lesson_3_Building_GIF_Grid.md
@@ -47,7 +47,8 @@ NICE. We have some spicy GIFs to test out our app with now. So, how are we going
 Let's start by creating a new function called `renderConnectedContainer`  right under the `renderNotConnectedContainer`. This will have some simple UI code that will map through all our GIF links and render them:
 
 ```jsx
-const renderConnectedContainer = () => (
+const renderConnectedContainer = () => {
+return (
   <div className="connected-container">
     <div className="gif-grid">
       {TEST_GIFS.map(gif => (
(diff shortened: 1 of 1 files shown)


Solana_And_Web3/en/Section_1/Lesson_3_Building_GIF_Grid.md
```diff
@@ -268,7 +272,8 @@ All we are saying here is if we have a `walletAddress` go ahead and run our fetc
 So, once our test data is set, we want to use it! For this we are going to head back to the `renderConnectedContainer` function and make a one line change:
 
 ```jsx
-const renderConnectedContainer = () => (
+const renderConnectedContainer = () => {
```
helloitsm3, line 275: This is not needed. `=>` is already a return statement

Solana_And_Web3/en/Section_1/Lesson_3_Building_GIF_Grid.md
```diff
@@ -118,7 +120,8 @@ It's time for us to think about how other people can add their own GIFs to our a
 Let's get started with the input. We only want this input box to be shown when the user has connected their wallet to our app. So, that means we will want to add this code to our `renderConnectedContainer` render method:
 
 ```jsx
-const renderConnectedContainer = () => (
+const renderConnectedContainer = () => {
```
helloitsm3, line 123: This is not needed. => is already a return statement. Can you revert this?

Solana_And_Web3/en/Section_1/Lesson_3_Building_GIF_Grid.md
```diff
@@ -47,7 +47,8 @@ NICE. We have some spicy GIFs to test out our app with now. So, how are we going
 Let's start by creating a new function called `renderConnectedContainer`  right under the `renderNotConnectedContainer`. This will have some simple UI code that will map through all our GIF links and render them:
 
 ```jsx
-const renderConnectedContainer = () => (
+const renderConnectedContainer = () => {
```
helloitsm3, line 50: This is not needed. => is already a return statement. Can you revert this?

PR #1223, opened 2022-07-18
fixed repetition

lodgendury:
instruction to create api folder was already given previously.


AlmostEfficient (approved):

AlmostEfficient:
What would I do without you mate?

PR #1244, opened 2022-08-08
replace base64 lib.

b4s36t4:
Instead of adding a new file and copy pasting contents, as from openzepplin v4 we have base64 available, using that solidity file might be feasible.


helloitsm3 (approved):
LGTM!

AlmostEfficient (approved):

PR #1247, opened 2022-08-11
Update from Rinkeby to Goerli

aitor-react:
Changed a couple words to switch the deployment from Rinnkeby to Goerli as it's deprecating in October.
I saw there was already the same PR on Lesson_2_Deploy_To_Real_Testnet.md so with this one Section 2 would be complete :)


helloitsm3:
We've already migrated to Goerli

PR #1254, opened 2022-08-28
Hardhat-waffle to Hardhat-toolbox Migration Update

Rohitchaudhary13:
Hardhat has migrated from waffle to toolbox recently, which creates a lot of fuss for newbies, and also it is required to be always updated with the docs. So, in order to solve the issues, the documentation is required to contain the latest version of the hardhat with toolbox installation. I have updated the document precisely and checked for all possible issues. With this documentation, the hardhat is at its latest version now and works perfectly with all the upcoming docs.


helloitsm3:
Closing this. We've already updated the content

PR #1257, opened 2022-08-31
Fix: space units used for an account in example

benyam7:
This PR tries to resolve an issue with units used for the `space` of an account. I might be completely wrong but yeah 😅 

I think the unit for the `space` of an `account` is measured in `bytes`, not in `kilobytes`.  So, I think instead of this line `space = 8 + 8 (eg, 8 kilobytes + 8 kilobytes)` it should be `space = 8 + 8 (eg, 8 bytes + 8 bytes)`


helloitsm3 (approved):
LGTM!

AlmostEfficient (approved):

PR #1270, opened 2022-09-15
Update outdated information in M1 setup info doc

jennnx:
Hey! I posted in Discord about this, and wanted to suggest a PR too.

There's basically two things:
1) `anchor test` now runs out of the box. The developers there fixed this issue in  https://github.com/solana-labs/solana/pull/21215, where they automatically disabled `bpf-jit` in validators for m1s and windows. I tried to update that everywhere it was referenced.

2) I think the section on `solana-keygen -o` and outputting it to the `target/deploy` folder *might* be outdated. Maybe it was M1-specific, but following along with the main tutorial (the one for Linux/x86) where you just use `solana-keygen new` and use your default created wallet was way easier. In fact, since the `Anchor.toml` file generated by `anchor init` by default expects the wallet to be at `"/Users/x/.config/solana/id.json"`, it was giving me some errors before I did the switch.

Maybe it's better to just delete this setup-flow completely and make everyone use the main tutorial on the buildspace website! The only "broken" thing for the M1 side is just the actual installation of anchor itself (the main tutorial on buildspace pulls the CLI binary directly from anchor's github, but as per the docs m1 processors should use avm instead).

### DIFF
--- a/Solana_And_Web3/en/Section_2/Resources/m1_setup.md
+++ b/Solana_And_Web3/en/Section_2/Resources/m1_setup.md
@@ -166,17 +166,17 @@ The last thing to test is we want to make sure we can get a local Solana node ru
 
 First, shoutout to **@dimfled#9450**, and send some love his way! He has 'seen things' building with Solana recently and gave us this step to get things working on M1s!
 
-We are going to run our Solana validator manually. Will explain why we need this shortly:
+Let's now run our Solana validator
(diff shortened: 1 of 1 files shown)


AlmostEfficient:
Thanks a bunch for this, looking into it now!

AlmostEfficient (approved):

PR #1271, opened 2022-09-16
Migrated Intro-to-ETH to Goerli

helloitsm3:


AlmostEfficient:
Looks like this PR also has all the NFT changes, could you update to only have Goerli changes? Easier to revert if something goes wrong.

PR #1286, opened 2022-09-30
Update Lesson_1_Get_Local_Ethereum_Running.md

adanb13:
Does not work without the migration, as seen here: https://hardhat.org/hardhat-chai-matchers/docs/migrate-from-waffle


AlmostEfficient:
Tysm, merged in another PR!

PR #1294, opened 2022-10-14
Update Lesson_2_Call_Deployed_Program.md

scheuclu:
"AnchorProvider" is not exported by the package, thus throws an error. Changing it to "Provider" solved the issue.


helloitsm3:
We've already upgraded anchor to the latest version so this is not needed. Please ensure that your anchor is using version 0.25.0. Thanks

PR #1319, opened 2022-11-05
change _app.tsx to AppBar.tsx

matiman:
change _app.tsx to AppBar.tsx


AlmostEfficient:
I dunno if this is right - `_app.tsx` is the root component so by importing the wallet libs in there all child components get it

PR #1344, opened 2022-11-09
fixed semi-colon error

Rojeenyus:
if you use prettier for auto formatter, it will error if there is no semi-colon. saw other people having this issue also in discord


helloitsm3 (approved):
LGTM!

PR #1345, opened 2022-11-09
Update Lesson_11_Candy_Machine_And_The_Suger_CLI.md

scheuclu:
I had struggled a bit here, because I did not fully understand where the files go, and that they actually need to have numeric names. So I am proposing to make it a bit more clear.


helloitsm3 (approved):
LGTM

PR #1351, opened 2022-11-11
Lesson 14 - metaplex nft fetching by owner

archae0pteryx:
Fixes a bug in example with fetching of nft data.

- add `.run()` call for metaplex api
- response checks in place for data without uris
- refactors / small code improvements


AlmostEfficient:
Are you sure this works for the included setup instrucitons? `.run()` was deprecated in a recent version

PR #1354, opened 2022-11-12
Fixing command mismatch

albincsergo:
Changed not working `npm run dev` to working `npm start`


helloitsm3 (commented):
The starter template uses next.js so this command is valid

albincsergo:
> The starter template uses next.js so this command is valid
It didn't work for me, but I'm probably wrong. If that's the case sorry for wasting your time.

PR #1373, opened 2022-11-16
fix: Remove unused param in core1, lesson 4

jimthedev:
It turns out the connection isn't needed.

It is fixed in the solution already, but the docs need to be updated too. 

This PR updates the docs.


AlmostEfficient (approved):

AlmostEfficient:
Oooooooooh looks like I missed this, ty Jim!

PR #1377, opened 2022-11-17
Update Lesson_4_Build_An_Interaction_Script.md

RoberVH:
The connection parameter doesn't belong there. I posted the error is giving and  @bellu pointed out  the mistake. See here: https://discord.com/channels/869321815833575434/1032052070209040504/1042633421148663809

### DIFF
--- a/Solana_Core/en/Core_1/Lesson_4_Build_An_Interaction_Script.md
+++ b/Solana_Core/en/Core_1/Lesson_4_Build_An_Interaction_Script.md
@@ -69,7 +69,7 @@ Update your main function and run the script with `npm start` to test it out:
 ```ts
 async function main() {
   const connection = new Web3.Connection(Web3.clusterApiUrl('devnet'));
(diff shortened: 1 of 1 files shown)


IvoBiaus:
oh i saw the same, but i actually came with the opposite fix https://github.com/buildspace/buildspace-projects/pull/1376. I requested the parameter in the `initializeKeypair` function because you will need it a few steps later when you create the `airdropSolIfNeeded` function.

RoberVH:
> oh i saw the same, but i actually came with the opposite fix #1376. I requested the parameter in the `initializeKeypair` function because you will need it a few steps later when you create the `airdropSolIfNeeded` function.

That's right, I saw it later, for the sake of the lesson a note about this could be made to add that connection parameter later on the lesson and thus allowing the first time to run it smoothly

helloitsm3:
Thanks for this PR but it's already fixed in #1376

PR #1379, opened 2022-11-18
Update Lesson_4_Build_An_Interaction_Script.md

shungo0222:
The `initializeKeypair` function doesn't take any arguments.

### DIFF
--- a/Solana_Core/en/Core_1/Lesson_4_Build_An_Interaction_Script.md
+++ b/Solana_Core/en/Core_1/Lesson_4_Build_An_Interaction_Script.md
@@ -69,7 +69,7 @@ Update your main function and run the script with `npm start` to test it out:
 ```ts
 async function main() {
   const connection = new Web3.Connection(Web3.clusterApiUrl('devnet'));
-  const signer = await initializeKeypair(connection);
+  const signer = await initializeKeypair();
 
   console.log("Public key:", signer.publicKey.toBase58());
 }


helloitsm3 (commented):
LGTM!

helloitsm3 (approved):

IvoBiaus:
@helloitsm3 With this modification you left the next instruction without a parameter. 
(I mean when they have to call `airdropSolIfNeeded` inside the `initializeKeypair` function)

The correct way of fixing this error is what Robert and I were discussing in his PR https://github.com/buildspace/buildspace-projects/pull/1377
1. You either remove that parameter (like you did here) but also, a few lines later you would have to add instructions for the almuni about adding it again (because they will need it), you would also have to tell them to send that parameter's value when they call it in the `main` function.

2. The other option was what i meant with my PR https://github.com/buildspace/buildspace-projects/pull/1376.
Just accepting that parameter, even if they dont use it right away, so they have it ready for when the next instruction comes.

helloitsm3:
> 2\. or w

Yup, your PR works too. Will merge yours.

IvoBiaus:
> > 2. or w
> 
> Yup, your PR works too. Will merge yours.

Ok well, but in that case i think we should rollback the change made here, right? Do you want me to create a PR with that?
Because with this PR's change we are not sending the `connection` in the function call.

helloitsm3:
> > > 2. or w
> > 
> > 
> > Yup, your PR works too. Will merge yours.
> 
> Ok well, but in that case i think we should rollback the change made here, right? Do you want me to create a PR with that? Because with this PR's change we are not sending the `connection` in the function call.

Yes please make another PR for this change

shungo0222:
@helloitsm3 @IvoBiaus 
Thank you for your trouble.

PR #1385, opened 2022-11-19
Update Lesson_16_Create_Reward_Tokens.md

shungo0222:
* delete `findMetadataPda` and use `metaplex.nfts().pdas().metadata()` because it's deprecated
* remove `run()` because you don't need to use it in the latest version
* `ts-node ./tokens/bld/index.ts` -> `ts-node ./src/tokens/bld/index.ts` because there is src dir above tokens dir.
Screen Shot 2022-11-19 at 17 33 59
Screen Shot 2022-11-19 at 17 34 43


helloitsm3 (approved):
This is epic! Thank you so much @shungo0222 💖

PR #1387, opened 2022-11-21
Defer referencing `connected` yet until it defined in the scope

samnang:
When I work through the content, I noticed there is a reference to `connected` variable before it hasn't yet been defined in the scope until a few sections further in the lesson. We should better remove the references and keep the app running build successful and reference it in the same section we define it.

```js
{!connected && <Disconnected />} // connected is not defined yet
```
Until few sections later, it is defined here:

```js
const { connected } = useWallet()

...

{connected ? <Connected /> : <Disconnected />}
```

### DIFF
--- a/Solana_Core/en/Core_1/Lesson_11_Build_NFT_Minter_Frontend.md
+++ b/Solana_Core/en/Core_1/Lesson_11_Build_NFT_Minter_Frontend.md
@@ -1,9 +1,9 @@
 ### 💻 Build an NFT minter front-end
-Welcome to your first week of SHIPPING. Every week you will have an entire section dedicated to taking your learnings and building it into your custom NFT staking app w/ loot boxes! 
+Welcome to your first week of SHIPPING. Every week you will have an entire section dedicated to taking your learnings and building it into your custom NFT staking app w/ loot boxes!
 
 The whole point of these sections is to get you off localhost and building something real that others can use. All the builders that have come before you have found wild success from just putting their work out there and building in public. This is the moment you have been preparing for -- **let's do this thing 🤘.**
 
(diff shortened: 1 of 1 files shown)


helloitsm3:
Thank you so much for this. However, since it's going to get implemented in the later section, do you think it's better to have it included here first instead of removing the variable?

samnang:
It might be okay to leave it there if we keep guiding people keep following step by step without mentioning the running server yet, but I went through the content and mentioned checking the server, I saw the error and a bit confused because I didn't know where `connected` came from.

image

helloitsm3:
> It might be okay to leave it there if we keep guiding people keep following step by step without mentioning the running server yet, but I went through the content and mentioned checking the server, I saw the error and a bit confused because I didn't know where `connected` came from.
> 
> image

You're right about this. It can be confusing. Can I trouble you to remove that line that says "Look at localhost:3000" and add an additional note to mention that they will see an error now but they can fix it in the later section? Thanks

samnang:
> You're right about this. It can be confusing. Can I trouble you to remove that line that says "Look at localhost:3000" and add an additional note to mention that they will see an error now but they can fix it in the later section? Thanks

When I looked at places that referenced `connected` again, I saw two places. From my suggestion, keeping the flow with working version of the app might be better for readers. Even if we tried to say it is okay to see the error about `connected`, readers might be confused because we never explain why it is needed. Finally, if we look at the screenshot below which is where it is mentioned and defined, it explains clearly why it's needed, so we don't see much needed to introduce this variable early in the top section.

image

helloitsm3 (approved):
LGTM

PR #1400, opened 2022-11-25
Remove async useEffect

ElliotWood13:
This syntax is not recommended in React and displays a linting error. I've changed as little as possible to keep what you intended but to also keep this tutorial as simple as possible.

Keep up the great work!


helloitsm3 (approved):
Good catch. Thank you so much for this 👍

PR #1402, opened 2022-11-25
"below" should be more accurate.

NemesisLW:
"Go ahead and write this function right *under* our `connectWallet()` function." - might seem like the wave function has to be inside the connectWallet() function, instead of it being a seperate function.


helloitsm3 (approved):
LGTM

PR #1408, opened 2022-11-28
Replace <insert> with real publishing guide

aleemrehmtulla:
currently: "you can read up on it here." with no link
proposed: same thing, link to https://developer.chrome.com/docs/webstore/publish/


helloitsm3 (approved):
LGTM!

PR #1417, opened 2022-11-28
Update Lesson_3_Connect_Wallet_To_Web_App.md

develo-pera:
The part which says "If we're logged in to Metamask, it will automatically inject a special object" is not entirely correct. If we have Metamask installed `ethereum` object will be injected whether we are logged in into it or not. It won't be injected only if we don't have it installed.


helloitsm3 (approved):
LGTM!

PR #1420, opened 2022-11-29
"Below" should be more accurate.

NemesisLW:
Again, "Under" - may seem like it should be inside the connectWallet() function, instead it being a seperate function.


helloitsm3 (approved):
LGTM!

PR #1421, opened 2022-11-29
Fixed Network Name in comments.

NemesisLW:
As we are connecting to goerli chain. changed the comment on line - 80 from Rinkebey to Goerli


helloitsm3 (approved):
LGTM!

PR #1422, opened 2022-11-29
Remove backticks from around link GPT3-section2

rhyspowell:
Rather than display the code display it as a link


helloitsm3 (approved):
LGTM!

PR #1427, opened 2022-11-30
Update Lesson_1_Randomly_Pick_Winner.md

develo-pera:
Change `Randomly pick winner` bug in code as it checks for `if (seed < 50)` and description to truly check for less than or equal, not just less than

### DIFF
--- a/Solidity_And_Smart_Contracts/en/Section_4/Lesson_1_Randomly_Pick_Winner.md
+++ b/Solidity_And_Smart_Contracts/en/Section_4/Lesson_1_Randomly_Pick_Winner.md
@@ -65,7 +65,7 @@ contract WavePortal {
         /*
          * Give a 50% chance that the user wins the prize.
          */
-        if (seed < 50) {
+        if (seed <= 50) {
             console.log("%s won!", msg.sender);
 
             /*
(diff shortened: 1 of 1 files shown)


helloitsm3:
This doesn't sound right. Please look at the explanation here on why it's < 50%. Do feel free to reopen the PR if you feel that it needs to be re-evaluated. Thanks

develo-pera:
Hey @helloitsm3, thanks for the reply. I do think it should be re-evaulated. Code doesn't lie :)

Descriptions says `Then I just write a simple if statement to see if the seed is less than or equal to 50` which is not true. Statement `if (seed < 50)` checks only if the seed is less than 50, not less or equal to 50. It's clear as a day. Saying we are now writing check that will have 50-50 (equal) chances is a different thing than saying that statement `if (seed < 50)` checks for less than or equal to 50.

Btw, you can also see that in the the same module in a section below that check is actually written as `if (seed <= 50)`.

I see two options:
1. Change description `Then I just write a simple if statement to see if the seed is less than or equal to 50` and all checks to be `if (seed < 50)`
2. Or go with this change that is already proposed 

What do you think?

helloitsm3:
Sorry for this. I might have missed out on that. Merging now

helloitsm3 (approved):

helloitsm3:
@all-contributors please add @develo-pera for docs

PR #1436, opened 2022-12-02
Bernard

dongdongzhuzhu:
translate section 0 of GPT3 write to chinese


dongdongzhuzhu:
translate section 0 of GPT3 write to chinese，If the quality is OK, I will continue to translate the following chapters

helloitsm3 (approved):
Good job. This is really good. Please continue translating the other lessons. Thank you

PR #1459, opened 2022-12-07
FIX issue 1458 - [Solana_And_Web3]: <SECTION_2> - <Lesson_2_Call_Deployed_Program>

Wonderland2077:
https://github.com/buildspace/buildspace-projects/issues/1458#issue-1481761606 
Which is about AnchorProvider not exists, due to the anchor version in package.json is not new enough.


helloitsm3 (approved):
Thank you so much for this

PR #1463, opened 2022-12-09
translate GPT_writer section2 into chinese

dongdongzhuzhu:


dongdongzhuzhu:
translate GPT_writer section3 into chinese

helloitsm3 (approved):

PR #1466, opened 2022-12-09
Added quotes

### LINKED ISSUE #1464: ✨ [JS_DAO]: <SECTION_#2> - <Lesson_3_Let_Users_Mint_Your_NFT>
### What You Are Changing
*A short description of the change you made and why!*
We need to add the editiondrop address in double quotes

image

Atharv181:
Closes #1464 
Without Double quotes contract got error and button was not loading as soon as I added the quotes it works completely fine.


helloitsm3:
You're suppose to replace `editionDropAddress` with your own Edition Drop Address

PR #1497, opened 2022-12-18
Update Lesson_13_Build_NFT_Staking_Program.md

shungo0222:
I think this is supposed to be "Staked" because this is in the "process_stake" function.


helloitsm3 (approved):
Good catch. Thanks for this

helloitsm3:
@all-contributors please add @shungo0222 for docs

PR #1499, opened 2022-12-20
Added cron job to remove stale PRs and issues

helloitsm3:
Workflow should automatically label PRs and issues as stale if there is no activity within 30 days and remove it after 14 days of no activity.


AlmostEfficient (approved):

helloitsm3:
> Lgtm, I'm wondering about older translation PRs - can we please tag them so they're kept track of and so that when we have ppl that can validate we can find them?

I'll look into seeing how we can automate this

AlmostEfficient (approved):

PR #1504, opened 2022-12-21
Update Lesson_13_Build_NFT_Staking_Program.md

s6thgehr:
Stake state should be StakeState::Staked after process_stake()


s6thgehr:
There is already a pull request for this issue

PR #1512, opened 2022-12-23
Update Lesson_2_Meta_learning_Bow_to_AI.md

neddinn:
Looks like there's a dangling prompt erroneously left there.


helloitsm3 (commented):
Why is this wrong? 🤔

neddinn:
> Why is this wrong? 🤔

In this example of `single-shot learning`, a `prompt:` `story:` mode is used. To quote just the next line below this, 
`I explicitly use Prompt: and Story: just so GPT-3 knows where the last story stopped, and where the next one should start.`

In this case, though, there is a prompt there without use because using it before the `prompt:` makes it a part of the previous story, and that's how I can tell it was left there erroneously maybe while copying from playground?

alexisweill:
Agreed with this change - I created a similar pr before realizing you already had open one so closing mine (https://github.com/buildspace/buildspace-projects/pull/1517)

helloitsm3:
Thanks for this. Will merge it

helloitsm3 (approved):

helloitsm3:
@all-contributors please add 
@neddinn for docs

PR #1551, opened 2023-01-07
Rename heading from "... React app" to "... Next.js app"

haytchde:
Not using React


helloitsm3:
Thanks for creating this PR but I don't think it's necessary to rename the file since Next.js is a React framework

PR #1552, opened 2023-01-07
Update prompt to convert the essay to a twitter thread.

BowTiedHeron:
Chat GPT was able to understand my request, but for playground I had to add a bit more instructions. 
I've submitted my prompt, as a suggestion :), to get the outcome below:

Take the essay below and turn it into a tweet thread; number the tweets and show breaks in between.

``` 

I don't think it's always about "solving a problem", sometimes it's worth thinking about creating something different. A lot of times, the best way to make an impact is by introducing something that's never been seen before. Take the example of Y-Combinator, the investment firm that I co-founded. We weren't trying to solve a problem as much as we were trying to create something that had never been done before. We were trying to build a new way for early-stage startups to get the funding and support they needed to succeed. We didn't have a roadmap laid out in front of us, and there were no guarantees that it would work. But we decided to take a risk and forge ahead anyway. We believed that if we could provide a new way of helping startups succeed, then the whole tech industry would benefit. And it worked. Y-Combinator has changed the game for early-stage startups. We're now able to give more startups the resources they need to succeed, and they, in turn, have been able to make an impact on the world. The lesson here is that sometimes, it's worth taking a risk and trying something new. You don't always have to find a solution to a problem – sometimes it's worth creating something new. And when you do, you can make a huge impact. 

```

1. I don't think it's always about "solving a problem", sometimes it's worth thinking about creating something different. #NewPerspective 

2. Take the example of Y-Combinator, the investment firm that I co-founded. We weren't trying to solve a problem as much as we were trying to create something that had never been done before. 

3. We were trying to build a new way for early-stage startups to get the funding and support they needed to succeed. We didn't have a roadmap laid out in front of us, and there were no guarantees that it would work. 

4. But we decided to take a risk and forge ahead anyway. We believed that if we could provide a new way of helping startups succeed, then the whole tech industry would benefit.

5. And it worked! Y-Combinator has changed the game for early-stage startups. We're now able to give more startups the resources they need to succeed, and they, in turn, have been able to make an impact on the world. 

6. The lesson here is that sometimes, it's worth taking a risk and trying something new. You don't always have to find a solution

### DIFF
--- a/GPT3_Writer/en/Section_1/Lesson_5_Secret_advanced_gpt3_stuff.md
+++ b/GPT3_Writer/en/Section_1/Lesson_5_Secret_advanced_gpt3_stuff.md
@@ -40,7 +40,7 @@ Let me show you something wild. Open a new tab with an empty Playground. So, you
 In Playground #2, go ahead and write this prompt:
 
 ```
-Take the essay below and turn it into a tweet thread.
+Take the essay below and turn it into a tweet thread; number the tweets and show breaks in between.
 ```
 
 Now, underneath there — copy paste the output from Playground #1. So for me it looks like this:


helloitsm3 (approved):

helloitsm3:
@all-contributors please add @BowTiedHeron for docs

PR #1555, opened 2023-01-07
Fixed focusing out of currentWindow issue.

NemesisLW:
used the `url` parameter to point to the text-editor directly, `chrome.tabs.query({ url: "URL_of_your_text_editor" }, (tabs) => { const activeTab = tabs[0].id;` instead of the `chrome.tabs.query({ active: true, currentWindow: true }`.


helloitsm3:
Is this tested and ensure that it's working? What happens if your browser can't find the tab with that url? Will it throw any errors? @NemesisLW

PR #1557, opened 2023-01-08
Patch depreciated Body.buffer() method in Avatar Generator

ElliotPadfield:
`.buffer()` is a depreciated method from earlier versions of node/node-fetch causing method not found errors. `.arrayBuffer()` is suitable for all versions. Have swapped out the body data conversion to ArrayBuffer & modified `bufferToBase64` function to correctly convert the ArrayBuffer to Base64 for display.


helloitsm3 (approved):
Thanks for this <3

PR #1560, opened 2023-01-09
Update 'unique subject identifier' subsection

AryanMadhavVerma:
Add line of code to also change body parameter from `input` to `finalInput` in api/generate.js file along with index.js file

### DIFF
--- a/AI_Avatar_Generator/en/Section_4/Lesson_1_The_finishing_touches.md
+++ b/AI_Avatar_Generator/en/Section_4/Lesson_1_The_finishing_touches.md
@@ -15,6 +15,10 @@ const response = await fetch('/api/generate', {
   body: JSON.stringify({ input: finalInput }),
 });
 ```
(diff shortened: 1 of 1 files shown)


AI_Avatar_Generator/en/Section_4/Lesson_1_The_finishing_touches.md
```diff
@@ -15,6 +15,10 @@ const response = await fetch('/api/generate', {
   body: JSON.stringify({ input: finalInput }),
 });
 ```
+Make sure to change `input` to `finalInput` in api/generate.js file as well!
+```jsx
+const input = JSON.parse(req.body).finalInput; //body has finalInput as a parameter now
```
helloitsm3, line 20:
```suggestion
const { finalInput } = JSON.parse(req.body); // body has finalInput as a parameter now
```

PR #960, opened 2022-02-21
fix: out of scope error for "ProgramResult"

ripunjaynarula:
When we run "anchor test" command, we get the error "not found in this scope" i.e. error [E0412] of rust language docs. the error has been fixed by adding a declaration :)

PR #1562, opened 2023-01-10
Remove unnecessary use

LStan:


helloitsm3:
Why was this removed? Was it not used in the program?

LStan:
ProgramResult was used as return type for methods and was changed to Result<()> in anchor 0.22.0 released on 2022-02-20. This `use` was added in [#960](https://github.com/buildspace/buildspace-projects/pull/960) on 2022-02-22, so probably the old release of anchor was used.
And in the next lessons there is no this `use`

helloitsm3:
> 

Thanks for catching this

helloitsm3 (approved):

PR #1567, opened 2023-01-12
Important update

anipaul2:
Really frustrating if the spl-token version is latest as npm run start doesn't work with the lastest version.

### DIFF
--- a/Solana_Core/en/Core_2/Lesson_3_Mint_Tokens_One_Solana.md
+++ b/Solana_Core/en/Core_2/Lesson_3_Mint_Tokens_One_Solana.md
@@ -133,7 +133,7 @@ async function main() {
 }
 ```
 
-Run `npm run start` - you should see three explorer links logged in the terminal. **Save the token mint account address.** You'll need it later. Open up the last link and scroll down to the token balances section:
+Run `npm run start` - you should see three explorer links logged in the terminal. (**Note**: Make sure you have @solana/spl-token@0.2.0 or it will show error. To install -> npm install @solana/spl-token@0.2.0 and change package.json dependencies to "@solana/spl-token": "^0.2.0",) **Save the token mint account address.** You'll need it later. Open up the last link and scroll down to the token balances section:
 
 ![](https://hackmd.io/_uploads/BJBT5nXms.png)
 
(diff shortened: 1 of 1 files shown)


Solana_Core/en/Core_2/Lesson_3_Mint_Tokens_One_Solana.md
```diff
@@ -133,7 +133,7 @@ async function main() {
 }
 ```
 
-Run `npm run start` - you should see three explorer links logged in the terminal. **Save the token mint account address.** You'll need it later. Open up the last link and scroll down to the token balances section:
+Run `npm run start` - you should see three explorer links logged in the terminal. (**Note**: Make sure you have @solana/spl-token@0.2.0 or it will show error. To install -> npm install @solana/spl-token@0.2.0 and change package.json dependencies to "@solana/spl-token": "^0.2.0",) **Save the token mint account address.** You'll need it later. Open up the last link and scroll down to the token balances section:
```
helloitsm3, line 136: Highlight the commands so that users knows which command to run. You don't have to manually change the package.json version as it will automatically change the version for you. 

```suggestion
Run `npm run start` - you should see three explorer links logged in the terminal. (**Note**: Make sure you have `@solana/spl-token@0.2.0` or it will show error. To install type `npm uninstall @solana/spl-token` and `npm install @solana/spl-token@0.2.0` into the terminal **Save the token mint account address.** You'll need it later. Open up the last link and scroll down to the token balances section:
```

anipaul2: > Highlight the commands so that users knows which command to run. You don't have to manually change the package.json version as it will automatically change the version for you.

Yeah but what if the person doesn't has lower verision? then package will add the installed version and npm run start or dev doesn't works.

helloitsm3: Please do not mark the comments as resolve. Should leave it to the reviewers to mark it. The reason why you don't have to manually change the version in the package.json is because you require the user to first uninstall the package and reinstall it in that specific version

helloitsm3 (approved):

PR #1579, opened 2023-01-20
Fixing inconsistency in course material text

ANMOLCODES:
Hi gang, noticed a possible type in the course material, corrected it!


helloitsm3 (commented):
ML stands for `Machine Learning` so `ML` makes sense in this context

PR #1581, opened 2023-01-21
Adding in a 'console.log' for secondPromptCompletion.text

vvronskyFX:
This'll help with getting a console output for the secondPrompt to make sure it's working properly.


helloitsm3:
Thanks for this but it is not necessary to have it in. If you feel that it should be re-reviewed, do feel free to re-open this PR.

PR #1592, opened 2023-02-05
Fix: add address that was undefined.

mi2valley:
Added `editionDropAddress` that was undefined (line 53).


helloitsm3 (commented):
This seems redundant. You can just replace `INSERT_EDITION_DROP_ADDRESS` with your actual address

mi2valley:
Thank you for your comment.
I think the original code would output an error due to an undefined `editionDropAddress` on Line 80.

helloitsm3:
> Thank you for your comment. I think the original code would output an error due to an undefined `editionDropAddress` on Line 80.

Thanks for this. Will merge

helloitsm3 (approved):

PR #1599, opened 2023-02-09
Update yarn commands to use 'yarn install'

clairefro:
`yarn add next...` is not necessary as the project starter comes with all the dependencies already listed in package.json. `yarn install` will do the trick


helloitsm3 (commented):
Why is this needed? `yarn install` should already be expected. The reason why we include `yarn add next react react-dom` is for those who happen to not have these packages included in their `package.json`.

PR #1600, opened 2023-02-13
Add missing content

B8N4S8N:
defined reject for getKey function
added console.log(baseCompletion.text)   to first prompt to remedy implied deletion
added console.log(secondPromptCompletion);  at the end of the lesson, to allow builders to complete this task and move on without confusion.

### DIFF
--- a/GPT3_Writer/en/Section_3/Lesson_2_Call_OpenAI_from_extension.md
+++ b/GPT3_Writer/en/Section_3/Lesson_2_Call_OpenAI_from_extension.md
@@ -182,6 +182,8 @@ const getKey = () => {
       if (result['openai-key']) {
         const decodedKey = atob(result['openai-key']);
         resolve(decodedKey);
+	} else {
+          reject('Key not found');
       }
     });
   });
@@ -225,6 +227,8 @@ const generateCompletionAction = async (info) => {
     const baseCompletion = await generate(
       `${basePromptPrefix}${selectionText}`
     );
+    console.log(baseCompletion.text)	
+
 
     // Add your second prompt here
     const secondPrompt = `
@@ -239,6 +243,7 @@ const generateCompletionAction = async (info) => {
 
     // Call your second prompt
     const secondPromptCompletion = await generate(secondPrompt);
+     console.log(secondPromptCompletion);
   } catch (error) {
     console.log(error);
   }


GPT3_Writer/en/Section_3/Lesson_2_Call_OpenAI_from_extension.md
```diff
@@ -182,6 +182,8 @@ const getKey = () => {
       if (result['openai-key']) {
         const decodedKey = atob(result['openai-key']);
         resolve(decodedKey);
+	} else {
```
helloitsm3, line 185: Your `else` statement is in the wrong place

GPT3_Writer/en/Section_3/Lesson_2_Call_OpenAI_from_extension.md
```diff
@@ -225,6 +227,8 @@ const generateCompletionAction = async (info) => {
     const baseCompletion = await generate(
       `${basePromptPrefix}${selectionText}`
     );
+    console.log(baseCompletion.text)
```
helloitsm3, line 230: What's the reason for adding this console.log?

B8N4S8N: It doesn't NEED to be there, but the current set up has the potential of leading to confusion on whether half the lesson was even done right, as this is pre injection.  As it stands, the requirement for the completion of this lesson is the output of one of the first tasks within the lesson.  When users get no response as to whether the second completion even worked, it can lead to the assumption they didn't do the second half of this lesson right.  With this added console.log, it shows users they have successfully completed all tasks within this lesson.

GPT3_Writer/en/Section_3/Lesson_2_Call_OpenAI_from_extension.md
B8N4S8N: I am filled with much confuse.  Can you help me understand what you mean?  
The syntax and structure W3schools shows is as follows:
`if (hour < 18) { 
  greeting = "Good day";
} else {
  greeting = "Good evening";
}`    
This is what I have done, as far as structure goes. I tried matching the format used within the buildspace lesson so I have added parenthesis as well as omitted the =, maybe this is where the issue is?  Not having the else, shows warnings of unused statements, as reject is declared yet never used. I tried adding the else inside the if statement and get errors. I try adding it outside the function, I have errors. I added it right there, and it works as intended.   I'd really appreciate any help understanding how that is in the wrong spot. I am pretty new to this.

genesiscz: It has one tab more as it should have 

```
if (true) {
   // something
} else {
   // something else
}
```

not 


```
if (true) {
   // something
    } else {
   // something else
}
```

PR #1601, opened 2023-02-13
Add useEffect hook for address

alymurtazamemon:
We should add here the `useEffect` hook for the address so that our web application can update the state accordingly whenever the user connects and disconnects the wallet.
```jsx
// Use the hooks thirdweb give us.
  const address = useAddress();
  console.log("👋 Address:", address);
  
  useEffect(() => {}, [address]);
```


helloitsm3:
This doesn't really make much sense. `useAddress` is already a hook which will trigger a re-render every time there's an update so there's no reason why you need to have a `useEffect` in here

PR #1602, opened 2023-02-13
Change Website Path

alymurtazamemon:
```diff
-https://goerli.etherscan.io/
+https://mumbai.polygonscan.com/
```

Because we are using the Mumbai Testnet, the website path should be this.


helloitsm3:
Thanks for the contributiong but since this project is in Goerli, there's no need for this.

PR #1610, opened 2023-02-17
Potential trouble when running `Anchor test` to build, deploy and test in one command

0xk2:
### The problem
I have a quite old mac so my Solana build is somewhat a frankenstein with some binary I built myself and some come from pre-built official version. This bring me to various problem when using `anchor test` to go through buid -> deploy -> test automatically.

### How I fix it
I find out that the best way to learn Anchor is starting with the whole process of `buid -> deploy -> test`.
Then be able to run my own local validator then use `anchor test --skip-local` is quite helpful.
I follow the instruction [here](https://lorisleiva.com/create-a-solana-dapp-from-scratch/getting-started-with-solana-and-anchor) with all the program ID configuration process and it works nicely.


helloitsm3:
Seems like you didn't rebase your commit. Please ensure that you're working on the latest copy. Thanks

PR #1645, opened 2023-06-05
Update Lesson_2_Add_in_OpenAI_API.md

koushikVss:
While the prompt is submitted and output is being generated, the "callgeneratedEndpoint" on button Click is still active to make another request on clicking the pinning loader. So we put that onClick event listener in a condition to prevent it from making othe request to api.


koushikVss:
Please fix this

koushikVss:
> While the prompt is submitted and output is being generated, the "callgeneratedEndpoint" on button Click is still active to make another request on clicking the pinning loader. So we put that onClick event listener in a condition to prevent it from making othe request to api.

